PR #11, opened 2023-11-21
Update index.md

cloutiertyler:
Fixes a margin for the figure that kinda breaks it on mobile.


cloutiertyler:
https://github.com/clockworklabs/spacetime-docs/assets/6432361/f2a0e0d5-8928-4ade-9fcc-4ce13b85dd9a

Link on home page appears to be broken.

PR #35, opened 2024-02-26
Dylan/deploy-docs-impr

dylanh724:
## About

Split the only "Testnet" deploy section into 3, fixed syntax errors, added clarity,added missing steps, clarified local hosting, ensured `publish` command examples added names, added instructions on how to regenerate local+testnet.

## Why?

The Deployment section previously only vaguely covered `testnet` with some deep issues and lacking connections. Going through a few `spacetime publish` without a name would also create confusion when the web dashboard was introduced, likely seeing at least 2 unnamed projects if going through both the quickstart and a verbose tutorial.

## Details

- Deploy section split:
  - BEFORE: [ Testnet ]
  - AFTER: [ Overview, Hosted, Self-Hosted ]
- In root doc overview, added a name to the publish command (to prevent confusion later, especially from the web dash)
- Fixed syntax err: The instructions to set a default server *added* a server, but the current version already has `local` + `testnet`  servers added.
- Added instructions how to add back `local` and `testnet` servers via CLI after repeatedly experiencing a bug that wipes the servers.
- The new _overview_ section covers the general flow and high-level explanations.


docs/deploying/index.md
```diff
@@ -0,0 +1,48 @@
+# Deploying Overview
+
+SpacetimeDB supports both hosted and self-hosted publishing in multiple ways. Below, we will:
+
+1. Generally introduce Identities.
+1. Generally introduce Servers.
+1Choose to proceed with either a [Hosted](/docs/deploying/hosted.md) or [Self-Hosted](/docs/deploying/self-hosted.md) deployment.
```
bfops, line 7: typo? should it be `1. Choose` or just `Choose`?

docs/index.md
```diff
@@ -100,23 +100,21 @@ SpacetimeDB was designed first and foremost as the backend for multiplayer Unity
 1. How do I get/install SpacetimeDB?
    Just install our command line tool and then upload your application to the cloud.
 
-1. How do I create a new database with SpacetimeDB?
+4. How do I create a new database with SpacetimeDB?
```
bfops, line 103: why do we switch to explicit numbering here?

docs/nav.js
```diff
@@ -9,14 +9,21 @@ function section(title) {
 const nav = {
     items: [
         section("Intro"),
-        page("Overview", "index", "index.md"),
+        page("Overview", "index", "index.md"), // TODO(BREAKING): For consistency & clarity, 'index' slug should be renamed 'intro'?
```
bfops, line 12: I think that's a good rename, not sure if there's any other fallout I'm not aware/thinking of..

nav.ts
```diff
@@ -25,16 +25,24 @@ function section(title: string): NavSection {
 const nav: Nav = {
   items: [
     section("Intro"),
-    page("Overview",        "index",            "index.md"),
-    page("Getting Started", "getting-started",  "getting-started.md"),
+    page("Overview",            "index",                  "index.md"), // TODO(BREAKING): For consistency & clarity, 'index' slug should be renamed 'intro'?
```
bfops, line 28: (ditto above)

bfops (changes requested):
this broadly looks great, thank you! Left some minor comments.

docs/nav.js
bfops: it sounds like the `index` slug might be treated specially - can we avoid changing this for now?

docs/deploying/index.md
dylanh724: > typo? should it be `1. Choose` or just `Choose`?

`1. Choose` - fixing >>

docs/index.md
dylanh724: > why do we switch to explicit numbering here?

Editor shenanigans, likely - fixing >>

dylanh724: > why do we switch to explicit numbering here?

image
Ah, I see now - there was a lone `5.` below a bash script that wouldn't unhinged from the implicit.

I'll 1~5 them explicitly.

PR #42, opened 2024-04-19
Fixed code examples in rust reference regarding insertion

Muthsera:
While looking through the documentation, I noticed some inconsistencies in the section about insertion in the Rust reference.
Hope you don't mind this small PR regarding that.


cloutiertyler (approved):
Thank you!

PR #46, opened 2024-05-01
fix: Docs build, pnpm, vm evaluate

PuruVJ:
This started as a script to run tsc and push but then yarn generated a shit-load of files due to being being a different version than original, so I took this opportunity to just use pnpm.

This also evaluates the nav.js file and makes it a simple object, which is easier to parse on the upcoming remix app

Edit: Pasted in prettierrc from spacetime-web, to keep the consistency across


cloutiertyler (approved):
LGTM if you've tested and it works

cloutiertyler:
This appears to break the existing clone functionality, so I'm reverting for now.

PR #47, opened 2024-05-01
chore: Prettier on entire repo

PuruVJ:
Now that we have a .prettierrc, better to format the repo once so any future tweaks dont cause entire diffs full of prettification


cloutiertyler (approved):
Excellent, thank you for doing this as a separate PR

PuruVJ:
Closing this as it could cause git conflicts in current site which is quite hard to fix. Will take care of this later in Config-less architecture and when the remix version is up

Issue #40, opened 2024-04-05
proposal:  Config-less docs structure


PuruVJ:
## Problem

Currently any and all docs have to be specified in nav.js file. This is extra redundancy.

## Solution

We get rid of nav.js. Split the documentation into sections with .md files being the page slugs, and meta.json for section title.

Structure:

```txt
- introduction
  - meta.json              <----- This is a json file with { title: "Introduction" } as content
  - overview.md.           <----- This is URL: `/docs/introduction/overview`
  - getting-started.md     <----- This is URL: `/docs/introduction/getting-started`
- deploying
  - meta.json
  - testnet.md
- unity
  - meta.json
  - part-1.md
  - part-2.md
```

This will require setting up redirects for the overview routes, (which is like a 10 line piece of code), but the advantage is that our URLs will become consistent. Currently some urls have 0, some have 1 slug, some have 2, some even have 3. This will make em all just 2 fixed slugs, which is easier to reason with.
![CleanShot 2024-04-05 at 21 15 39@2x](https://github.com/clockworklabs/spacetime-docs/assets/47742487/8428ebb0-a783-46d6-a64f-602eb71d1bd3)
![CleanShot 2024-04-05 at 21 15 51@2x](https://github.com/clockworklabs/spacetime-docs/assets/47742487/f23b5e88-e484-4671-84d3-eeb0b29ab72b)
![CleanShot 2024-04-05 at 21 16 03@2x](https://github.com/clockworklabs/spacetime-docs/assets/47742487/9bcc4f5e-171a-4677-a25c-0ab883b9d723)

### How about title of a chapter?

This will be defined in the frontmatter:

```md
---
title: Testnet
---

<<CONTENT>>
```

This lets us add future config right within the content(manual next or previous, tags, version separation, etc)

## Prerequisites

Need the remix rewrite to become the prod site, as rewriting now will make it incompatible with the current site.

# Final complete structure:

```txt
- 00-introduction        
  - 00-overview.md       
  - 10-getting-started.md
- 01-deploying
  - 00-testnet.md
- 02-unity-basic
  - 00-1.md
  - 10-2a-rust.md
  - 20-2a-c-sharp.md
  - 30-3.md
- 03-unity-advanced
  - 00-4.md
  - 10-5.md
- 04-modules
  - 00-overview.md
  - 10-rust-quickstart.md
  - 20-rust-reference.md
  - 30-c-sharp-quickstart.md
  - 40-c-sharp-reference.md
[...]

PR #51, opened 2024-05-02
fix: Unity tutorial slugs

PuruVJ:


bfops (approved):
LGTM thank you! I didn't realize we had two versions of this file :exploding_head:

PuruVJ:
Yep, I wanna work on https://github.com/clockworklabs/spacetime-docs/issues/40 so we can get rid of both

PR #57, opened 2024-05-25
poc: Configless

PuruVJ:
POC for configless docs. Contrary to the name, these are infact quite config heavy. But all configruation is now spread across the entire filesystem structure of the docs. The filesystem, the metadata in meta.json and at the top of .md files will inform the final URLs and navbar structure on the website.


content/docs/08-ws/00-overview.md
```diff
@@ -1,42 +1,45 @@
...
+| ------------------------------------- | ------------------------------------ |
+| `v1.bin.spacetimedb`                  | [Binary](#protocols-binary-protocol) |
+| `v1.text.spacetimedb`                 | [Text](#protocols-text-protocol)     |
 
 ### Binary Protocol
 
-The SpacetimeDB binary WebSocket protocol, `v1.bin.spacetimedb`, encodes messages using [ProtoBuf 3](https://protobuf.dev), and reducer and row data using [BSATN](/docs/bsatn).
+The SpacetimeDB binary WebSocket protocol, `v1.bin.spacetimedb`, encodes messages using [ProtoBuf 3](https://protobuf.dev), and reducer and row data using [BSATN](/docs/data-format/bsatn).
 
 The binary protocol's messages are defined in [`client_api.proto`](https://github.com/clockworklabs/SpacetimeDB/blob/master/crates/client-api-messages/protobuf/client_api.proto).
```
PuruVJ, line 29: This is broken as we don't use protobufs anymore. What should this be replaced with?

gefjon: Ah, we need to rewrite this whole page, pretty much. Perhaps we should replace it with a stub for now, and aim to rewrite it after all the WebSocket API changes we have planned.

PuruVJ: Noted. Will do so in a separate PR after this

cloutiertyler:
@PuruVJ Spreading the config throughout the filesystem was how it used to be. We changed that (to copy the Bun structure) because it was very error prone. I don't think the meta.json file makes this better.

PuruVJ:
What were the errors that could happen?

cloutiertyler:
People editing the docs were unaware of the equivalent of the `.json` files and the docs would get out of sync. Putting it in one place made that less likely, although did not eliminate it. I think configless is an improvement, but this is not configless IMO.

PR #61, opened 2024-06-06
Update docs for `#[spacetimedb(table)]`

bfops:


docs/unity/part-2b-c-sharp.md
```diff
@@ -30,7 +30,7 @@ using SpacetimeDB.Module;
 using static SpacetimeDB.Runtime;
 ```
 
-Then we are going to start by adding the global `Config` table. Right now it only contains the "message of the day" but it can be extended to store other configuration variables. This also uses a couple of macros, like `#[spacetimedb(table)]` which you can learn more about in our [C# module reference](/docs/modules/c-sharp). Simply put, this just tells SpacetimeDB to create a table which uses this struct as the schema for the table.
+Then we are going to start by adding the global `Config` table. Right now it only contains the "message of the day" but it can be extended to store other configuration variables. This also uses a couple of attributes, like `[SpacetimeDB.Table]` which you can learn more about in our [C# module reference](/docs/modules/c-sharp). Simply put, this just tells SpacetimeDB to create a table which uses this struct as the schema for the table.
```
bfops, line 33: this is somewhat unrelated, but notice this copypasta in my `grep` and corrected it

gefjon (approved):
Needs a corresponding SpacetimeDB PR to update the quickstart-chat module.

bfops:
wait.. was the `table(public)` already released or should I not have merged that? For some reason I felt sure our docs were currently incorrect.

gefjon:
It was not already released. I guess you should either revert until 0.10 goes out.

PR #68, opened 2024-06-18
Update docs for Consistent Filtering proposal

RReverser:
Updating docs for FilterBy methods across SDKs, adding docs for FindBy methods, and a couple of drive-by fixes for broken links.


docs/sdks/c-sharp/index.md
```diff
@@ -394,22 +396,32 @@ SpacetimeDBClient.instance.connect(/* ... */);
...
+    public static TABLE? FindBySender(COLUMNTYPE value);
 }
 
 }
 ```
 
-For each column of a table, `spacetime generate` generates a static method on the [table class](#class-table) to filter or seek subscribed rows where that column matches a requested value. These methods are named `filterBy{COLUMN}`, where `{COLUMN}` is the column name converted to `PascalCase`.
-
-The method's return type depends on the column's attributes:
+For each unique column of a table (those annotated `#[unique]` or `#[primarykey]`), `spacetime generate` generates a static method on the [table class](#class-table) to seek subscribed rows where that column matches a requested value.
```
gefjon, line 422:
```suggestion
For each unique column of a table (those annotated `#[unique]` or `#[primarykey]`), `spacetime generate` generates a static method on the [table class](#class-table) to seek a subscribed row where that column matches a requested value.
```

gefjon (approved):

PR #71, opened 2024-07-16
docs(70): SQL spec for 1.0

joshua-spacetime:
There are two main sections of the doc. The first describes the subset of sql that is supported by the subscription api. The next one describes the level of sql support provided by the http api.

### DIFF
--- a/docs/sql/index.md
+++ b/docs/sql/index.md
@@ -1,407 +1,482 @@
 # SQL Support
 
-SpacetimeDB supports a subset of SQL as a query language. Developers can evaluate SQL queries against a Spacetime database via the `spacetime sql` command-line tool and the [`/database/sql/:name_or_address POST` HTTP endpoint](/docs/http/database#databasesqlname_or_address-post). Client developers also write SQL queries when subscribing to events in the [WebSocket API](/docs/ws#subscribe) or via an SDK `subscribe` function.
+SpacetimeDB supports two subsets of SQL:
+One for queries issued through the cli or [http] api.
+Another for subscriptions issued via the [sdk] or [websocket] api.
 
-SpacetimeDB aims to support much of the [SQL 2016 standard](https://www.iso.org/standard/63555.html), and in particular aims to be compatible with [PostgreSQL](https://www.postgresql.org/).
+## Subscriptions
 
-SpacetimeDB 0.6 implements a relatively small subset of SQL. Future SpacetimeDB versions will implement additional SQL features.
-
-## Types
-
-| Type                                          | Description                            |
-| --------------------------------------------- | -------------------------------------- |
-| [Nullable types](#nullable-types)             | Types which may not hold a value.      |
-| [Logic types](#logic-types)                   | Booleans, i.e. `true` and `false`.     |
-| [Integer types](#integer-types)               | Numbers without fractional components. |
-| [Floating-point types](#floating-point-types) | Numbers with fractional components.    |
-| [Text types](#text-types)                     | UTF-8 encoded text.                    |
-
-### Definition statements
-
-| Statement                     | Description                          |
-| ----------------------------- | ------------------------------------ |
-| [CREATE TABLE](#create-table) | Create a new table.                  |
-| [DROP TABLE](#drop-table)     | Remove a table, discarding all rows. |
-
-### Query statements
-
-| Statement         | Description                                                                                  |
-| ----------------- | -------------------------------------------------------------------------------------------- |
-| [FROM](#from)     | A source of data, like a table or a value.                                                   |
-| [JOIN](#join)     | Combine several data sources.                                                                |
-| [SELECT](#select) | Select specific rows and columns from a data source, and optionally compute a derived value. |
-| [DELETE](#delete) | Delete specific rows from a table.                                                           |
-| [INSERT](#insert) | Insert rows into a table.                                                                    |
-| [UPDATE](#update) | Update specific rows in a table.                                                             |
-
-## Data types
-
-SpacetimeDB is built on the Spacetime Algebraic Type System, or SATS. SATS is a richer, more expressive type system than the one included in the SQL language.
-
-Because SATS is a richer type system than SQL, some SATS types cannot cleanly correspond to SQL types. In particular, the SpacetimeDB SQL interface is unable to construct or compare instances of product and sum types. As such, SpacetimeDB SQL must largely restrict themselves to interacting with columns of builtin types.
-
-Most SATS builtin types map cleanly to SQL types.
-
-### Nullable types
-
-SpacetimeDB types, by default, do not permit `NULL` as a value. Nullable types are encoded in SATS using a sum type which corresponds to [Rust's `Option`](https://doc.rust-lang.org/stable/std/option/enum.Option.html). In SQL, such types can be written by adding the constraint `NULL`, like `INT NULL`.
-
-### Logic types
-
-| SQL       | SATS   | Example         |
-| --------- | ------ | --------------- |
-| `BOOLEAN` | `Bool` | `true`, `false` |
-
-### Numeric types
-
-#### Integer types
-
-An integer is a number without a fractional component.
-
-Adding the `UNSIGNED` constraint to an integer type allows only positive values. This allows representing a larger positive range without increasing the width of the integer.
-
-| SQL                 | SATS  | Example | Min    | Max   |
-| ------------------- | ----- | ------- | ------ | ----- |
-| `TINYINT`           | `I8`  | 1       | -(2⁷)  | 2⁷-1  |
-| `TINYINT UNSIGNED`  | `U8`  | 1       | 0      | 2⁸-1  |
-| `SMALLINT`          | `I16` | 1       | -(2¹⁵) | 2¹⁵-1 |
-| `SMALLINT UNSIGNED` | `U16` | 1       | 0      | 2¹⁶-1 |
-| `INT`, `INTEGER`    | `I32` | 1       | -(2³¹) | 2³¹-1 |
-| `INT UNSIGNED`      | `U32` | 1       | 0      | 2³²-1 |
-| `BIGINT`            | `I64` | 1       | -(2⁶³) | 2⁶³-1 |
-| `BIGINT UNSIGNED`   | `U64` | 1       | 0      | 2⁶⁴-1 |
-
-#### Floating-point types
-
-SpacetimeDB supports single- and double-precision [binary IEEE-754 floats](https://en.wikipedia.org/wiki/IEEE_754).
-
-| SQL               | SATS  | Example | Min                      | Max                     |
-| ----------------- | ----- | ------- | ------------------------ | ----------------------- |
-| `REAL`            | `F32` | 1.0     | -3.40282347E+38          | 3.40282347E+38          |
-| `DOUBLE`, `FLOAT` | `F64` | 1.0     | -1.7976931348623157E+308 | 1.7976931348623157E+308 |
-
-### Text types
-
-SpacetimeDB supports a single string type, `String`. SpacetimeDB strings are UTF-8 encoded.
-
-| SQL                                             | SATS     | Example | Notes                |
-| ----------------------------------------------- | -------- | ------- | -------------------- |
-| `CHAR`, `VARCHAR`, `NVARCHAR`, `TEXT`, `STRING` | `String` | 'hello' | Always UTF-8 encoded |
-
-> SpacetimeDB SQL currently does not support length contraints like `CHAR(10)`.
-
-## Syntax
-
-### Comments
-
-SQL line comments begin with `--`.
-
-```sql
--- This is a comment
+```ebnf
+SELECT projection FROM relation [ WHERE predicate ]
 ```
 
-### Expressions
+The subscription language is strictly a query language.
+Its sole purpose is to replicate a subset of the rows in the database,
+and to **automatically** update them in realtime as the database changes.
 
-We can express different, composable, values that are universally called `expressions`.
+There is no context for manually updating this view.
+Hence data manipulation commands like `INSERT` and `DELETE` are not supported.
 
-An expression is one of the following:
+> NOTE: Because subscriptions are evaluated in realtime,
+> performance is critical, and as a result,
+> additional restrictions are applied over ad hoc queries.
+> These restrictions are highlighted below.
 
-#### Literals
+### SELECT
 
-| Example   | Description |
-| --------- | ----------- |
-| `1`       | An integer. |
-| `1.0`     | A float.    |
-| `'hello'` | A string.   |
-| `true`    | A boolean.  |
+```ebnf
+SELECT ( '*' | table '.' '*' )
+```
 
-#### Binary operators
+The `SELECT` clause determines the table that is being subscribed to.
+Since the subscription api is purely a replication api,
+a query may only return rows from a single table,
+and it must return the entire row.
+Individual column projections are not allowed.
 
-| Example | Description         |
-| ------- | ------------------- |
-| `1 > 2` | Integer comparison. |
-| `1 + 2` | Integer addition.   |
-
-#### Logical expressions
-
-Any expression which returns a boolean, i.e. `true` or `false`, is a logical expression.
-
-| Example          | Description                                                  |
-| ---------------- | ------------------------------------------------------------ |
-| `1 > 2`          | Integer comparison.                                          |
-| `1 + 2 == 3`     | Equality comparison between a constant and a computed value. |
-| `true AND false` | Boolean and.                                                 |
-| `true OR false`  | Boolean or.                                                  |
-| `NOT true`       | Boolean inverse.                                             |
-
-#### Function calls
-
-| Example         | Description                                        |
-| --------------- | -------------------------------------------------- |
-| `lower('JOHN')` | Apply the function `lower` to the string `'JOHN'`. |
-
-#### Table identifiers
-
-| Example       | Description               |
-| ------------- | ------------------------- |
-| `inventory`   | Refers to a table.        |
-| `"inventory"` | Refers to the same table. |
-
-#### Column references
-
-| Example                    | Description                                             |
-| -------------------------- | ------------------------------------------------------- |
-| `inventory_id`             | Refers to a column.                                     |
-| `"inventory_id"`           | Refers to the same column.                              |
-| `"inventory.inventory_id"` | Refers to the same column, explicitly naming its table. |
-
-#### Wildcards
-
-Special "star" expressions which select all the columns of a table.
-
-| Example       | Description                                             |
-| ------------- | ------------------------------------------------------- |
-| `*`           | Refers to all columns of a table identified by context. |
-| `inventory.*` | Refers to all columns of the `inventory` table.         |
-
-#### Parenthesized expressions
-
-Sub-expressions can be enclosed in parentheses for grouping and to override operator precedence.
-
-| Example       | Description             |
-| ------------- | ----------------------- |
-| `1 + (2 / 3)` | One plus a fraction.    |
-| `(1 + 2) / 3` | A sum divided by three. |
-
-### `CREATE TABLE`
-
-A `CREATE TABLE` statement creates a new, initially empty table in the database.
-
-The syntax of the `CREATE TABLE` statement is:
-
-> **CREATE TABLE** _table_name_ (_column_name_ _data_type_, ...);
-
-![create-table](/images/syntax/create_table.svg)
+A `*` projection is allowed when the table is unambiguous,
+otherwise it must be qualified with the appropriate table name.
 
 #### Examples
 
-Create a table `inventory` with two columns, an integer `inventory_id` and a string `name`:
-
 ```sql
-CREATE TABLE inventory (inventory_id INTEGER, name TEXT);
+-- Subscribe to all rows of a table
+SELECT * FROM Inventory
+
+-- Qualify the `*` projection with the table
+SELECT item.* from Inventory item
+
+-- Subscribe to all customers who have orders totaling more than $1000
+SELECT customer.*
+FROM Customers customer JOIN Orders o ON customer.id = o.customer_id
+WHERE o.amount > 1000
+
+-- INVALID: Must return `Customers` or `Orders`, but not both
+SELECT *
+FROM Customers customer JOIN Orders o ON customer.id = o.customer_id
+WHERE o.amount > 1000
 ```
 
-Create a table `player` with two integer columns, an `entity_id` and an `inventory_id`:
+### FROM
 
-```sql
-CREATE TABLE player (entity_id INTEGER, inventory_id INTEGER);
+```ebnf
+FROM table [ [AS] alias ] [ [INNER] JOIN table [ [AS] alias ] ON column '=' column ]
 ```
 
-Create a table `location` with three columns, an integer `entity_id` and floats `x` and `z`:
+While you can only subscribe to rows from a single table,
+you may reference two tables in the `FROM` clause using a `JOIN`.
+A `JOIN` selects all combinations of rows from its input tables,
+and `ON` determines which combinations are considered.
+
+Subscriptions do not support joins of more than two tables.
+
+For any column referenced in `ON` clause of a `JOIN`,
+it must be qualified with the appropriate table name or alias.
+
+In order for a `JOIN` to be evaluated efficiently,
+subscriptions require an index to be defined on both join columns.
+
+#### Example
 
 ```sql
-CREATE TABLE location (entity_id INTEGER, x REAL, z REAL);
+-- Subscribe to all orders of products with less than 10 items in stock.
+-- Must have an index on the `product_id` column of the `Orders` table,
+-- as well as the `id` column of the `Product` table.
+SELECT o.*
+FROM Orders o JOIN Inventory product ON o.product_id = product.id
+WHERE product.quantity < 10
+
+-- Subscribe to all products that have at least one purchase
+SELECT product.*
+FROM Orders o JOIN Inventory product ON o.product_id = product.id
+
+-- INVALID: Must qualify the column names referenced in `ON`
+SELECT product.* FROM Orders JOIN Inventory product ON product_id = id
 ```
 
-### `DROP TABLE`
+### WHERE
 
-A `DROP TABLE` statement removes a table from the database, deleting all its associated rows, indexes, constraints and sequences.
+```ebnf
+predicate
+    = expr
+    | predicate AND predicate
+    | predicate OR  predicate
+    ;
 
-To empty a table of rows without destroying the table, use [`DELETE`](#delete).
+expr
+    = literal
+    | column
+    | expr op expr
+    ;
 
-The syntax of the `DROP TABLE` statement is:
+op
+    = '='
+    | '<'
+    | '>'
+    | '<' '='
+    | '>' '='
+    | '!' '='
+    | '<' '>'
+    ;
 
-> **DROP TABLE** _table_name_;
-
-![drop-table](/images/syntax/drop_table.svg)
-
-Examples:
-
-```sql
-DROP TABLE inventory;
+literal
+    = INTEGER
+    | STRING
+    | HEX
+    | TRUE
+    | FALSE
+    ;
 ```
 
-## Queries
+While the `SELECT` clause determines the table,
+the `WHERE` clause determines the rows in the subscription.
 
-### `FROM`
-
-A `FROM` clause derives a data source from a table name.
-
-The syntax of the `FROM` clause is:
-
-> **FROM** _table_name_ _join_clause_?;
-
-![from](/images/syntax/from.svg)
+Arithmetic expressions are not supported.
 
 #### Examples
 
-Select all rows from the `inventory` table:
-
 ```sql
-SELECT * FROM inventory;
+-- Find products that sell for more than $X
+SELECT * FROM Inventory WHERE price > {X}
+
+-- Find products that sell for more than $X and have fewer than Y items in stock
+SELECT * FROM Inventory WHERE price > {X} AND amount < {Y}
 ```
 
-### `JOIN`
+## Query and DML (Data Manipulation Language)
 
-A `JOIN` clause combines two data sources into a new data source.
+### Statements
 
-Currently, SpacetimeDB SQL supports only inner joins, which return rows from two data sources where the values of two columns match.
+- [SELECT](#select-1)
+- [INSERT](#insert)
+- [DELETE](#delete)
+- [UPDATE](#update)
+- [SET](#set)
+- [SHOW](#show)
 
-The syntax of the `JOIN` clause is:
+### SELECT
 
-> **JOIN** _table_name_ **ON** _expr_ = _expr_;
-
-![join](/images/syntax/join.svg)
-
-### Examples
-
-Select all players rows who have a corresponding location:
-
-```sql
-SELECT player.* FROM player
- JOIN location
- ON location.entity_id = player.entity_id;
+```ebnf
+SELECT projection FROM relation [ WHERE predicate ]
 ```
 
-Select all inventories which have a corresponding player, and where that player has a corresponding location:
+The query languge is a strict superset of the subscription language.
+The main differences are seen in column projections and [joins](#from-clause).
 
-```sql
-SELECT inventory.* FROM inventory
- JOIN player
- ON inventory.inventory_id = player.inventory_id
- JOIN location
- ON player.entity_id = location.entity_id;
+The subscription api only supports `*` projections,
+but the query api supports individual column projections.
+
+The subscription api limits the number of tables you can join,
+and enforces index constraints on the join columns,
+but the query language has no such constraints or limitations.
+
+#### SELECT Clause
+
+```ebnf
+projection
+    = '*'
+    | table '.' '*'
+    | projExpr { ',' projExpr }
+    ;
+
+projExpr
+    = column [ [ AS ] alias ]
+    ;
 ```
 
-### `SELECT`
+The `SELECT` clause determines the columns that are returned.
 
-A `SELECT` statement returns values of particular columns from a data source, optionally filtering the data source to include only rows which satisfy a `WHERE` predicate.
+##### Examples
 
-The syntax of the `SELECT` command is:
+```sql
+-- Select the items in my inventory
+SELECT * FROM Inventory;
 
-> **SELECT** _column_expr_ > **FROM** _from_expr_
-> {**WHERE** _expr_}?
+-- Select the names and prices of the items in my inventory
+SELECT item_name, price FROM Inventory
+```
 
-![sql-select](/images/syntax/select.svg)
+#### FROM Clause
+
+```ebnf
+FROM table [ [AS] alias ] { [INNER] JOIN table [ [AS] alias ] ON predicate }
+```
+
+Unlike [subscriptions](#from), the query api supports joining more than two tables.
+
+##### Examples
+
+```sql
+-- Find all customers who ordered a particular product and when they ordered it
+SELECT customer.first_name, customer.last_name, o.date
+FROM Customers customer
+JOIN Orders o ON customer.id = o.customer_id
+JOIN Inventory product ON o.product_id = product.id
+WHERE product.name = {product_name}
+```
+
+#### WHERE Clause
+
+See [Subscriptions](#where).
+
+### INSERT
+
+```ebnf
+INSERT INTO table [ '(' column { ',' column } ')' ] VALUES '(' literal { ',' literal } ')'
+```
 
 #### Examples
 
-Select all columns of all rows from the `inventory` table:
-
 ```sql
-SELECT * FROM inventory;
-SELECT inventory.* FROM inventory;
-```
+-- Inserting one row
+INSERT INTO Inventory (item_id, item_name) VALUES (1, 'health1');
 
-Select only the `inventory_id` column of all rows from the `inventory` table:
-
-```sql
-SELECT inventory_id FROM inventory;
-SELECT inventory.inventory_id FROM inventory;
-```
-
-An optional `WHERE` clause can be added to filter the data source using a [logical expression](#logical-expressions). The `SELECT` will return only the rows from the data source for which the expression returns `true`.
-
-#### Examples
-
-Select all columns of all rows from the `inventory` table, with a filter that is always true:
-
-```sql
-SELECT * FROM inventory WHERE 1 = 1;
-```
-
-Select all columns of all rows from the `inventory` table with the `inventory_id` 1:
-
-```sql
-SELECT * FROM inventory WHERE inventory_id = 1;
-```
-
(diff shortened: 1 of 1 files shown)


docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+aggrExpr
+    = COUNT '(' STAR ')' [AS] alias
+    | COUNT '(' DISTINCT column ')' [AS] alias
+    | SUM   '(' column ')' [AS] alias
+    ;
+```
+
+The `SELECT` clause determines the columns that are returned.
+In particular it supports both column and wildcard projections.
+It also supports sum and count aggregation.
```
mamcx, line 123:
```suggestion
It also supports `sum` and `count` aggregation.
```

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
+# SQL Support
+
+SpacetimeDB supports two subsets of SQL.
+One for ad hoc queries against the database.
+These can be made through the [cli] or the [http] endpoint.
+Another for subscriptions made through the [websocket] api or [sdk].
+
+## Subscriptions
+
```
mamcx, line 9: I think is a good idea to preface this by saying that subscriptions require precise semantics for build-derived data and that our SQL subset is designed to perform well and avoid issues.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+```ebnf
+SELECT projection FROM relation [ WHERE predicate ]
+```
+
+The subscription language is a query language.
+It does not support data manipulation commands.
```
mamcx, line 15: Why not?

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+It is used to subscribe to a set of rows from a single table.
```
mamcx, line 16: Below it says we can use joins, so sounds contradictory.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+
+```ebnf
+SELECT ( '*' | table '.' '*' )
+```
+
+The `SELECT` clause determines the table that is being subscribed to.
+A wildcard projection `*` is allowed when the table is unambiguous.
+A qualified wildcard projection `.*` is needed when there are multiple tables.
+Column projections are not allowed in subscriptions.
+
```
mamcx, line 28: Is incorrect `select * from t1 join t2`?  because the language below make me think this should say:



>A wildcard projection `*` is allowed only if is a single table, else you need to do a qualified wildcard projection `.*

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+FROM table [ [AS] alias ] [ [INNER] JOIN table [ [AS] alias ] ON predicate ]
+```
+
+While you can only subscribe to rows from a single table,
+you may reference two tables in the `FROM` clause using a `JOIN`.
+A `JOIN` selects all combinations of rows from its input tables,
+and `ON` determines which combinations are considered.
+Note that if a `JOIN` is used,
+an index must be defined on the join column of both tables,
+so that it can be evaluated efficiently.
```
mamcx, line 41:
```suggestion
> IMPORTANT: if a `JOIN` is used,
an index *must* be defined on the join column of both tables,
so that it can be evaluated efficiently.
```

so it  stand out

docs/sql/index.md
```diff
@@ -1,407 +0,0 @@
-# SQL Support
```
joshua-spacetime, line 1: Note, I deleted this file and replaced it with a new one for an easier to review diff. I'm not sure if this will render. I will rename back to `index.md` before merging.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+
+While the `SELECT` clause determines the table,
+the `WHERE` clause determines the rows in the subscription.
+
+The subscription language supports conjunctions `AND`.
+It does not support disjunctions `OR`.
+It does not support `!=` comparisons.
+It does not support arithmetic expressions.
+Note that disjunctions can be expressed by subscribing to multiple queries.
+
```
mamcx, line 82: Why not?

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+    ;
+```
+
+The `SELECT` clause determines the columns that are returned.
+In particular it supports both column and wildcard projections.
+It also supports sum and count aggregation.
+
+If `DISTINCT` is specified, only unique rows are returned.
+However each projected column must be of a type that allows comparison.
+
```
mamcx, line 127: Which types are not allowed? I think all of our `sats` types allow it, at least, all the `sql` ones that we support.

docs/sql/index.md
```diff
@@ -1,407 +0,0 @@
...
-
-### `CREATE TABLE`
-
-A `CREATE TABLE` statement creates a new, initially empty table in the database.
-
-The syntax of the `CREATE TABLE` statement is:
-
-> **CREATE TABLE** _table_name_ (_column_name_ _data_type_, ...);
-
-![create-table](/images/syntax/create_table.svg)
```
joshua-spacetime, line 184: Images point to the [spacetime-web](https://github.com/clockworklabs/spacetime-web) repo. Which means they don't render here.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+Examples:
+```sql
+-- Select the items in my inventory
+SELECT * FROM Inventory;
+
```
mamcx, line 132: Maybe is too much, but could be nice to show the output of the queries...

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+#### FROM Clause
+
+```ebnf
+FROM table [ [AS] alias ] { [INNER] JOIN table [ [AS] alias ] ON predicate }
+```
+
+The query language supports joins among an arbitrary number of tables.
+It does not impose any index restrictions like the subscription language.
```
mamcx, line 147:
```suggestion
> NOTE: It does not impose any index restrictions like the subscription language.
```

So it stand out

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+SpacetimeDB supports two subsets of SQL.
+One for ad hoc queries against the database.
+These can be made through the [cli] or the [http] endpoint.
+Another for subscriptions made through the [websocket] api or [sdk].
+
+## Subscriptions
+
+```ebnf
+SELECT projection FROM relation [ WHERE predicate ]
+```
```
joshua-spacetime, line 12: Note that previously the syntax was described using syntax diagrams stored in the [spacetime-web](https://github.com/clockworklabs/spacetime-web) repo. The `ebnf` specification makes this document easier to review, but probably less consumable from a UX perspective.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+    | STRING
+    | HEX
+    | TRUE
+    | FALSE
+    ;
+```
+
+The query language supports both conjunctions and disjunctions.
+It also supports `!=` comparisons.
+It does not support arithmetic expressions.
```
mamcx, line 195:
```suggestion
 > TBD: Support arithmetic expressions
```

I think we should clarify that certain features will be added later, no that should not be used.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+`ORDER BY` sorts the result set by one or more of the projected columns.
+Each column involved must be of a type that allows comparison.
+Ascending is the default behavior if not qualified.
+
+Examples:
+```sql
+SELECT * FROM Inventory ORDER BY price;
+```
```
mamcx, line 215:
```suggestion
SELECT * FROM Inventory ORDER BY price ASC;
SELECT * FROM Inventory ORDER BY price DESC
```
```

File: docs/sql/sql.md [hunk 16]
```diff
@@ -0,0 +1,440 @@
...
+
+Updates column values of existing rows in a table.
+The columns are identified by the `assignment` defined as `column '=' expr`.
+The column values are updated for all rows that match the `WHERE` condition.
+The rows are updated after the `WHERE` condition is evaluated for all rows.
+
+Examples:
+```sql
+-- Update the item_name for all rows with a specific item_id
```
mamcx, line 277:
```suggestion
-- Update all the rows
UPDATE Inventory SET item_name = 'new name';
-- Update the item_name for all rows with a specific item_id
```

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+Updates the value of a system variable.
+
+### SHOW
+
+```ebnf
+SHOW var
+```
+
```
mamcx, line 294: Which system variables we have?

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+Returns the value of a system variable.
+
```
mamcx, line 296: I think the above could be answered talking about the system tables.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+FLOAT
+    = [ '+' | '-' ] [ NUM ] '.' NUM
+    | [ '+' | '-' ] [ NUM ] '.' NUM 'E' [ '+' | '-' ] NUM
+    ;
+```
+
+### String
+
+The construction of variable length string values in Spacetime SQL,
+where `CHAR` is a `utf-8` encoded unicode character.
```
mamcx, line 362: mmm, so this should be `varchar` instead.

docs/sql/sql.md
```diff
@@ -0,0 +1,440 @@
...
+
+NUM
+    = DIGIT { DIGIT }
+    ;
+
+DIGIT
+    = 0..9
+    ;
+```
+
```
mamcx, line 346: I think we need to add a mapping here, and elsewhere when type inference is done, meaning that in fact we support `TINYINT, INTEGER, BIGINT, ` etc


| Sats Type    | SQL Type     | Description                                   |
|--------------|--------------|-----------------------------------------------|
| `i8`         | `TINYINT`    | 8-bit signed integer                          |
| `u8`         | `UNSIGNED TINYINT` | 8-bit unsigned integer                |
| `i16`        | `SMALLINT`   | 16-bit signed integer                         |
| `u16`        | `UNSIGNED SMALLINT` | 16-bit unsigned integer              |
| `i32`        | `INT`        | 32-bit signed integer                         |
| `u32`        | `UNSIGNED INT` | 32-bit unsigned integer                  |
| `i64`        | `BIGINT`     | 64-bit signed integer                         |
| `u64`        | `UNSIGNED BIGINT` | 64-bit unsigned integer                |
| `i128`       | Not directly supported | 128-bit signed integer (handled via custom logic or string in SQL) |
| `u128`       | Not directly supported | 128-bit unsigned integer (handled via custom logic or string in SQL) |

docs/sql/sql.md
joshua-spacetime: Yes `select * from t1 join t2` is invalid.

docs/sql/sql.md
joshua-spacetime: Are you asking for yourself or as a user? Are you saying that we should provide explanations for these restrictions?

docs/sql/sql.md
joshua-spacetime: This is the difference between SATS data types and the Spacetime SQL type system.

Take sum types for example. They have a binary representation, BSATN, with **a** total ordering, but that ordering is not unique, nor is it fixed. In fact the module is not even required to use it. The module could define its own ordering, or none at all.

So until orderings are tracked by the database, Spacetime SQL should be conservative in what it allows.

docs/sql/sql.md
joshua-spacetime: Perhaps a better place for that would be the module/subscription api docs since that defines the output format.

docs/sql/sql.md
joshua-spacetime: Good question. I wasn't sure if this was the correct place to specify them. We also haven't formally decided which ones we want to stabilize for 1.0.

docs/sql/sql.md
joshua-spacetime: Are you referring to the SATS String type?

docs/sql/sql.md
joshua-spacetime: I'm not quite sure what this means? Do you mean the system variables can be specified alongside the documentation for the system tables?

docs/sql/sql.md
joshua-spacetime: SATS specifies all of this, so I didn't want to duplicate. Instead I just refer to SATS. This is mainly to keep the docs concise and reduce bloat.

docs/sql/sql.md
joshua-spacetime: Is it made clear as you read on, or do you think more clarification is needed?

docs/sql/sql.md
mamcx: I'm asking as a user. I think we should explain when the restriction is fundamental or temporal, and why.

docs/sql/sql.md
mamcx: I mean we can show the system tables and that query them answer this kind of questions.

docs/sql/sql.md
mamcx: No, `sql` type.

docs/sql/sql.md
mamcx: Maybe is more clear if says:

The subscription language is strictly a query language designed specifically for subscribing to a set of rows from a single table, so is not allowed to do data manipulation commands.

docs/sql/sql.md
joshua-spacetime: I'll leave this one for now, just because updating all rows is less common. But it is specified in the language.

docs/sql/sql.md
joshua-spacetime: There was some confusion here. `CHAR` is in reference to the ebnf snippet right below this line. Not the SQL type.

docs/sql/sql.md
joshua-spacetime: Clarifying note: Mario notes this table was important because we used to interpret these SQL types in `CREATE TABLE`. Now that we no longer support DDL, it's not as important to present this mapping.

docs/sql/sql.md
joshua-spacetime: I will make mention of this.

docs/sql/sql.md
```diff
@@ -0,0 +1,475 @@
...
+While you can only subscribe to rows from a single table,
+you may reference two tables in the `FROM` clause using a `JOIN`.
+A `JOIN` selects all combinations of rows from its input tables,
+and `ON` determines which combinations are considered.
+
+> IMPORTANT: If a `JOIN` is used,
+> an index **must** be defined on the join column of both tables,
+> so that it can be evaluated efficiently.
```
joshua-spacetime, line 54: Note this applies to subscriptions only. This just repeats what is already specified in the proposal, but I want to highlight it in case it was overlooked.

We could make the same restriction for ad hoc queries. Mainly to avoid performance footguns. Ultimately these potential performance issues should be addressed by auth and/or workload management, but we won't have a complete solution that addresses them for 1.0. So I'm bringing this up here in case there are concerns.

docs/sql/sql.md
```diff
@@ -0,0 +1,475 @@
...
+UPDATE Inventory SET item_name = 'new name' WHERE item_id = 1;
+```
+
+### SET
+
+```ebnf
+SET var ( TO | '=' ) literal
+```
+
+Updates the value of a system variable.
```
joshua-spacetime, line 322: System variables are stored in the `st_var` table. Should these variables be specified in this document or somewhere else?

Note that we haven't formally stabilized any of the system variables for 1.0. Right now we have `row_limit` and a few others for logging slow queries.

So my questions are:

- Should they be part of the stable api?
    - If so, which ones?
    - If not, does it make sense for `SET var` and `SHOW var` to be part of the stable sql api?

cloutiertyler: Could we mark them as unstable in the documentation? I would be okay to do that as long as we make it clear to users that they are unstable.

I would lump SET and SHOW as part of the unstable API if all of the variables are also unstable.

docs/sql/sql.md
cloutiertyler: I would specify in the docs note that it is for subscriptions only.

For 1.0 we will have proper energy metering as it's required for Mainnet. I don't think we should disallow it for ad hoc queries as it may be required for things like allowing the BitCraft team to debug the data for example.

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+#### ORDER BY Clause
+
+```ebnf
+ORDER BY column [ ASC | DESC ] { ',' column [ ASC | DESC ] }
+```
+
+`ORDER BY` sorts the result set by one or more of the projected columns.
+Ascending is the default behavior if not qualified.
+Each column involved must be of a type that allows comparison.
+Comparison is not supported for sum, product, or array types.
```
mamcx, line 243:
```suggestion
> NOTE: Comparison is not supported for `sum`, `product`, or `array` types.
```

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+
+LATIN
+    = a..z | A..Z
+    ;
+```
+
+Identifiers are tokens that identify database objects like tables or columns.
+Spacetime SQL supports both quoted and unquoted identifiers.
+Quoted identifiers are case sensitive,
+whereas unquoted identifiers are case insensitive.
```
mamcx, line 460: This means that identifiers become lowercase like in PG.?

mamcx (commented):
I think this LGTM, but wait for a native review.

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
+# SQL Support
+
+SpacetimeDB supports two subsets of SQL.
+One for ad hoc queries against the database.
+These can be made through the [cli] or the [http] endpoint.
```
cloutiertyler, line 5:
```suggestion
(These can be made through the [cli] or the [http] endpoint.)
```

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
+# SQL Support
+
+SpacetimeDB supports two subsets of SQL.
```
cloutiertyler, line 3:
```suggestion
SpacetimeDB supports two subsets of SQL:
```

docs/sql/sql.md
cloutiertyler: I actually think it might be better UX. As an anecdote, I find these easier to reason about.

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+```ebnf
+SELECT projection FROM relation [ WHERE predicate ]
+```
+
+The subscription language is strictly a query language.
+Its sole purpose is to replicate a subset or view of the database,
+and to **automatically** update that view in realtime as the database changes.
+
+There is no context for manually updating this view.
+Hence data manipulation commands are not supported.
```
cloutiertyler, line 19:
```suggestion
Hence data manipulation commands, like [example], are not supported.
```

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+    | TRUE
+    | FALSE
+    ;
+```
+
+While the `SELECT` clause determines the table,
+the `WHERE` clause determines the rows in the subscription.
+
+The subscription language supports conjunctions `AND`.
+It does not support disjunctions `OR`.
```
cloutiertyler, line 91:
```suggestion
It does not yet support disjunctions `OR`.
```

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+It does not support `!=` comparisons.
```
cloutiertyler, line 92:
```suggestion
It does not yet support `!=` comparisons.
```

docs/sql/sql.md
cloutiertyler: Is this a yet or ever situation?

docs/sql/sql.md
cloutiertyler: Is this a yet or ever situation?

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+SELECT * FROM Inventory ORDER BY price;
+
+-- Sort the items in the inventory by order of decreasing price
+SELECT * FROM Inventory ORDER BY price DESC
+
+#### LIMIT Clause
+
+```ebnf
+LIMIT INTEGER
+```
```
cloutiertyler, line 257: Should we define an unsigned `INTEGER` type?

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+If `WHERE` is specified, only the matching rows are deleted.
+
+Examples:
+```sql
+-- Delete all rows
+DELETE FROM Inventory;
+
+-- Delete all rows with a specific item_id
+DELETE FROM Inventory WHERE item_id = 1;
+```
```
cloutiertyler, line 298: Love the examples!

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+
+Examples:
+```sql
+-- Update the item_name for all rows with a specific item_id
+UPDATE Inventory SET item_name = 'new name' WHERE item_id = 1;
+```
+
+### SET
+
+> NOTE: This statement is not part of the stable API.
```
cloutiertyler, line 319:
```suggestion
> WARNING: This statement is not part of the stable API.
```

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+
+```ebnf
+SET var ( TO | '=' ) literal
+```
+
+Updates the value of a system variable.
+
+### SHOW
+
+> NOTE: This statement is not part of the stable API.
```
cloutiertyler, line 330:
```suggestion
> WARNING: This statement is not part of the stable API.
```

docs/sql/sql.md
```diff
@@ -0,0 +1,498 @@
...
+
+```ebnf
+SHOW var
+```
+
+Returns the value of a system variable.
+
+## System Variables
+
+> NOTE: System variables are not part of the stable API.
```
cloutiertyler, line 341:
```suggestion
> WARNING: System variables are not part of the stable API.
```

docs/sql/sql.md
cloutiertyler: Is there any issue with just making them all case sensitive? I assume that's not part of the SQL standard, but boy does it avoid some confusion.

cloutiertyler (approved):
Approved with various nits mentioned throughout. The only thing I would consider more deeply is whether unquoted identifiers really need to be case insensitive.

docs/sql/sql.md
joshua-spacetime: That's the only issue. We just wouldn't be SQL Standard compliant.

docs/sql/sql.md
joshua-spacetime: We could.

docs/sql/sql.md
joshua-spacetime: Yet.

docs/sql/sql.md
joshua-spacetime: Probably yet, but with details.

docs/sql/sql.md
joshua-spacetime: Thank you @mamcx!

docs/sql/sql.md
joshua-spacetime: It's your call. Should we include both?

docs/sql/sql.md
joshua-spacetime: Another option would be to nix unquoted identifiers altogether. This would make it so that everything is case sensitive, but we still maintain compliance with the standard. It's not a great UX though, since all identifiers, even column names, will need to be quoted.

joshua-spacetime: Again it's your call.

docs/sql/sql.md
cloutiertyler: Just keep `ebnf` only.

docs/sql/sql.md
cloutiertyler: Let's stick with the standard. So how you have it written.

joshua-spacetime:
I'm waiting to merge this until, at the very least, the spec as been fully implemented, since merging to master automatically deploys doc changes.

joshua-spacetime:
NOTE: There are some sql tickets scheduled for 1.0, so I'm moving this to 1.0. It can be merged to master anytime after 0.12 is released. It won't be deployed until it's incorporated into the release branch.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+## Query and DML
 
-#### Binary operators
+### Statements
 
-| Example | Description         |
-| ------- | ------------------- |
-| `1 > 2` | Integer comparison. |
-| `1 + 2` | Integer addition.   |
+- [SELECT](#select)
```
gefjon, line 116: I don't think this link goes where you want it to; I believe it will find the first `#select` anchor, which is the above one about subscriptions, not the below one about one-off queries.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+```ebnf
+SELECT projection FROM relation [ WHERE predicate ]
+```
 
-The syntax of the `CREATE TABLE` statement is:
+The query languge is a strict superset of the subscription language.
+The main differences are seen in column projections and joins.
 
-> **CREATE TABLE** _table_name_ (_column_name_ _data_type_, ...);
+The subscription api only supports wildcard projections,
```
gefjon, line 132: If we're going to talk about "wildcard projections" this much, I think we should define that term somewhere. It would also be nice to say *why* subscriptions only support wildcard queries.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+but the query api supports individual column projections.
 
-![create-table](/images/syntax/create_table.svg)
+The subscription api limits the number of tables you can join,
+and enforces index constraints on the join columns,
+but the query language has no such constraints or limitations.
```
gefjon, line 137: This could use an example too. Show me a 3-way join, or something.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
-```
+-- Select the items in my inventory
+SELECT * FROM Inventory;
 
-Create a table `location` with three columns, an integer `entity_id` and floats `x` and `z`:
-
-```sql
-CREATE TABLE location (entity_id INTEGER, x REAL, z REAL);
+-- Select the names of the items in my inventory
+SELECT item_name FROM Inventory
```
gefjon, line 161: Can you add an example that selects more than one but fewer than `*` columns?

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
-To empty a table of rows without destroying the table, use [`DELETE`](#delete).
-
-The syntax of the `DROP TABLE` statement is:
-
-> **DROP TABLE** _table_name_;
+#### FROM Clause
 
-![drop-table](/images/syntax/drop_table.svg)
+```ebnf
+FROM table [ [AS] alias ] { [INNER] JOIN table [ [AS] alias ] ON predicate }
```
gefjon, line 167: I get that `JOIN` is grammatically a child of `FROM`, but I wonder if we should give it a separate section. We have to devote a lot of text to what joins we do and don't support, and I expect we'll want to link directly to that.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+-- Inserting two rows
+INSERT INTO Inventory (item_id, item_name) VALUES (1, 'health1'), (2, 'health2');
+```
 
-Currently, SpacetimeDB SQL supports only inner joins, which return rows from two data sources where the values of two columns match.
+### DELETE
 
-The syntax of the `JOIN` clause is:
+```ebnf
+DELETE FROM table [ WHERE predicate ]
```
gefjon, line 201: Do we allow joins as filters in deletes?

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
 
-Select all players rows who have a corresponding location:
+### UPDATE
 
-```sql
-SELECT player.* FROM player
- JOIN location
- ON location.entity_id = player.entity_id;
+```ebnf
+UPDATE table SET [ '(' assignment { ',' assignment } ')' ] [ WHERE predicate ]
```
gefjon, line 219: Same question re: joins.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
- ON player.entity_id = location.entity_id;
+-- Update the item_name for all rows with a specific item_id
+UPDATE Inventory SET item_name = 'new name' WHERE item_id = 1;
 ```
 
-### `SELECT`
+### SET
 
-A `SELECT` statement returns values of particular columns from a data source, optionally filtering the data source to include only rows which satisfy a `WHERE` predicate.
+> WARNING: This statement is not part of the stable API.
```
gefjon, line 235: Can we say "Support for the `SET` statement is experimental" instead? The appeal to "the stable API" reads weird to me.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+```
 
-![sql-select](/images/syntax/select.svg)
+Updates the value of a system variable.
 
-#### Examples
+### SHOW
 
-Select all columns of all rows from the `inventory` table:
+> WARNING: This statement is not part of the stable API.
```
gefjon, line 246: Again, "the stable API" feels like a silly thing to say. There's no platonic "stable API" sitting in the realm of ideas.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+Returns the value of a system variable.
 
-```sql
-SELECT inventory_id FROM inventory;
-SELECT inventory.inventory_id FROM inventory;
-```
+## System Variables
 
-An optional `WHERE` clause can be added to filter the data source using a [logical expression](#logical-expressions). The `SELECT` will return only the rows from the data source for which the expression returns `true`.
+> WARNING: System variables are not part of the stable API.
```
gefjon, line 257: ibid

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
 
-```sql
-SELECT name FROM inventory WHERE inventory_id = 1;
+```ebnf
+literal = INTEGER | FLOAT | STRING | HEX | TRUE | FALSE ;
 ```
 
-Select all columns of all rows from the `inventory` table where the `inventory_id` is 2 or greater:
+The following describes how to construct literal values,
+for [SATS] data types, in Spacetime SQL.
```
gefjon, line 285: Is `[SATS]` a meaningful link? I don't think it particularly adds value to this sentence, either way.

```suggestion
The following describes how to construct literal values of various data types in Spacetime SQL.
```

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
-> **INSERT INTO** _table_name_ (_column_name_, ...) **VALUES** (_expr_, ...), ...;
-
-![sql-insert](/images/syntax/insert.svg)
+DIGIT
+    = 0..9
+    ;
+```
 
-#### Examples
+> NOTE: The concrete [SATS] type is inferred from the context.
```
gefjon, line 308:
```suggestion
The concrete integer type is inferred from the context.
```

I don't think this needs special highlighting.

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
-INSERT INTO inventory (inventory_id, name) VALUES (1, 'health1');
+```ebnf
+FLOAT
+    = [ '+' | '-' ] [ NUM ] '.' NUM
+    | [ '+' | '-' ] [ NUM ] '.' NUM 'E' [ '+' | '-' ] NUM
+    ;
 ```
 
-Insert two rows:
+> NOTE: The concrete [SATS] type is inferred from the context.
```
gefjon, line 319:
```suggestion
The concrete floating-point type is inferred from the context.
```

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
 ```
 
-### DELETE
-
-A `DELETE` statement deletes rows that satisfy the `WHERE` clause from the specified table.
+Identifiers are tokens that identify database objects like tables or columns.
+Spacetime SQL supports both quoted and unquoted identifiers.
 
-If the `WHERE` clause is absent, the effect is to delete all rows in the table. In that case, the result is a valid empty table.
+> NOTE: Both types of identifiers are case sensitive.
```
gefjon, line 363:
```suggestion
Both types of identifiers are case sensitive.
```

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
+LATIN
+    = a..z | A..Z
+    ;
 ```
 
-### DELETE
-
-A `DELETE` statement deletes rows that satisfy the `WHERE` clause from the specified table.
+Identifiers are tokens that identify database objects like tables or columns.
+Spacetime SQL supports both quoted and unquoted identifiers.
```
gefjon, line 361: What's the difference between a quoted and an unquoted identifier? When would I use one or the other?

docs/sql/index.md
```diff
@@ -1,407 +1,391 @@
...
-Delete all rows from the `inventory` table, leaving it empty:
 
-```sql
-DELETE FROM inventory;
-```
+[sdk]:       /docs/sdks/rust/index.md#subscribe-to-queries
+[http]:      /docs/http/database#databasesqlname_or_address-post
+[websocket]: /docs/ws/index.md#subscribe
+
+[sats]:     /docs/satn.md
```
gefjon, line 389: Oh, you're sending `[SATS]` to the poorly-named `satn.md`, which describes our JSON encoding. I do not think that's a good idea.

gefjon (changes requested):

docs/sql/index.md
joshua-spacetime: No we do not. The grammar is explicit about this, but I have also added text to clarify.

docs/sql/index.md
joshua-spacetime: The link is not meaningful, but it is important to specify the type system under which these values are interpreted. For instance, it could be expected that numeric values have arbitrary precision which they do not.

docs/sql/index.md
joshua-spacetime: Yes, this was originally just a placeholder, but we still don't have a doc that describes the type system apart from a specific encoding. I removed the link for now, but ideally we would like to link to that reference doc.

docs/sql/index.md
joshua-spacetime: Logically it's part the `FROM` clause, so I've kept it like that, but I've added links to the appropriate section whenever I mention joins.

docs/sql/index.md
```diff
@@ -29,30 +29,69 @@ SELECT ( '*' | table '.' '*' )
...
+FROM table [ [AS] alias ] [ [INNER] JOIN table [ [AS] alias ] ON column '=' column ]
 ```
 
 While you can only subscribe to rows from a single table,
 you may reference two tables in the `FROM` clause using a `JOIN`.
 A `JOIN` selects all combinations of rows from its input tables,
 and `ON` determines which combinations are considered.
 
-> NOTE: Joining more than two tables will be supported in the future.
+Subscriptions do not yet support joins of more than two tables.
```
gefjon, line 71:
```suggestion
Subscriptions do not support joins of more than two tables.
```

docs/sql/index.md
```diff
@@ -29,30 +29,69 @@ SELECT ( '*' | table '.' '*' )
...
+
+Note that for any column referenced in `ON` clause of a `JOIN`,
```
gefjon, line 73:
```suggestion
For any column referenced in `ON` clause of a `JOIN`,
```

docs/sql/index.md
```diff
@@ -29,30 +29,69 @@ SELECT ( '*' | table '.' '*' )
...
+it must be qualified with the appropriate table name or alias.
+Also note that in order for a `JOIN` to be evaluated efficiently,
```
gefjon, line 75:
```suggestion

In order for a `JOIN` to be evaluated efficiently,
```

docs/sql/index.md
```diff
@@ -91,29 +130,23 @@ literal
 While the `SELECT` clause determines the table,
 the `WHERE` clause determines the rows in the subscription.
 
+> NOTE: Arithmetic expressions are not yet supported.
```
gefjon, line 133:
```suggestion
Arithmetic expressions are not supported.
```

docs/sql/index.md
```diff
@@ -204,7 +243,10 @@ DELETE FROM table [ WHERE predicate ]
 Deletes all rows from a table.
 If `WHERE` is specified, only the matching rows are deleted.
 
-Examples:
+> NOTE: `DELETE` does not support joins.
```
gefjon, line 246:
```suggestion
`DELETE` does not support joins.
```

docs/sql/index.md
```diff
@@ -224,16 +266,19 @@ The columns are identified by the `assignment` defined as `column '=' literal`.
 The column values are updated for all rows that match the `WHERE` condition.
 The rows are updated after the `WHERE` condition is evaluated for all rows.
 
-Examples:
+> NOTE: `UPDATE` does not support joins.
```
gefjon, line 269:
```suggestion
`UPDATE` does not support joins.
```

docs/sql/index.md
```diff
@@ -266,10 +311,10 @@ Returns the value of a system variable.
 
 ## Data types
 
-The data types that SpacetimeDB supports is defined by [SATS],
+The data types that SpacetimeDB supports is defined by SATS,
```
gefjon, line 314:
```suggestion
The data types that SpacetimeDB supports are defined by SATS,
```

Or possibly, "The set of data types [...] is defined [...]"

docs/sql/index.md
```diff
@@ -282,7 +327,7 @@ literal = INTEGER | FLOAT | STRING | HEX | TRUE | FALSE ;
 ```
 
 The following describes how to construct literal values,
-for [SATS] data types, in Spacetime SQL.
+for SATS data types, in Spacetime SQL.
```
gefjon, line 330: Commas aren't necessary here - "for SATS data types" is not a subordinate clause. GH won't let me suggest, but should be:

```
The following describes how to construct literal values for SATS data types in Spacetime SQL.
```

docs/sql/index.md
```diff
@@ -293,7 +338,7 @@ Booleans are represented using the canonical atoms `true` or `false`.
 ```ebnf
 INTEGER
     = [ '+' | '-' ] NUM
-    | [ '+' | '-' ] NUM 'E' [ '+' | '-' ] NUM
+    | [ '+' | '-' ] NUM 'E' [ '+' ] NUM
```
gefjon, line 341: Wait, we support scientific notation integers? I didn't notice this the first time through, but... huh.

gefjon (approved):
This is really great. I've left a few minor suggestions, mostly either shuffling commas around or removing the word "yet."

docs/sql/index.md
joshua-spacetime: https://github.com/clockworklabs/SpacetimeDB/pull/2196

PR #75, opened 2024-07-31
Update part-2b-c-sharp.md

rlneumiller:
The intent is to throw an exception if the player already exists, not the other way 'round.


cloutiertyler (approved):
Thank you!

PR #76, opened 2024-08-02
Shub/revert scheduler table doc

Shubham8287:


bfops (approved):
```bash
$ git diff origin/master~1 origin/shub/revert-scheduler-table-doc
```
this returns nothing, so I'm calling this a correct revert

PR #80, opened 2024-08-25
Update Rust Quickstart to use correct function to find User

Savalige:
Updated the Rust Quickstart documentation to use filter_by_identity instead of find_by_identity as done in the current Quickstart example.


cloutiertyler:
Thanks so much! We just found this issue simultaneously as well. Will approve!

cloutiertyler (approved):

PR #86, opened 2024-09-16
Fix broken tutorial package link

jdetter:
- We merged the C#/Unity SDKs recently which broke this part of the tutorial. The tutorial is still pretty broken but we're doing a pass on this soon and this is a good fix for now.


bfops (approved):
Sure, this is incrementally less broken

PR #93, opened 2024-10-02
Update Rust SDK ref for the new SDK

gefjon:
Docs are somewhat rushed and less complete than they used to be, but what are you gonna do.

(jdetter) Rendered: https://github.com/clockworklabs/spacetime-docs/blob/5f636ddedb6435e841c7da6f00c04d45fc24828e/docs/sdks/rust/index.md


jdetter (approved):
> Docs are somewhat rushed and less complete than they used to be, but what are you gonna do.

Yeah essentially the goal is:
1. Have at least some example code in the docs that works (this is supplied in the quickstarts)
2. Try not to have anything in the docs that is old or doesn't work

This PR I believe removes all references to anything that is outdated, looks good to me. Thanks Phoebe 🙏

PR #110, opened 2024-11-14
Style guide for our docs

gefjon:
It's a style guide! Which our docs should (but definitely currently do not) obey! We're gonna make everyone on the team read it!

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -93,6 +93,18 @@ Use first-person pronouns sparingly to draw attention to non-technical informati
 
 > We don't recommend that you use the `ArrayList` class for new development. Instead, we recommend that you use the generic `List<T>` class.
 
+#### *Usually* Avoid Passive Voice
+
+Use active voice rather than passive voice to avoid ambiguity regarding who is doing the action. Active voice directly attributes actions to the subject, making sentences easier to understand. For example:
+
+- Passive voice: "The method was invoked."
+- Active voice: "The user invoked the method."
+
+The second example is more straightforward and clarifies who is performing the action. In most cases, prefer using the active voice to maintain a clear and direct explanation of code behavior.
+
+However, passive voice may be appropriate in certain contexts where the actor is either unknown or irrelevant. In these cases, the emphasis is placed on the action or result rather than the subject performing it. For example:
+
+- "The `Dispose` method is called automatically when the object is garbage collected."
 ### Tables and links
 
 Each reference page should have one or more two-column tables, where the left column are namespace-qualified names or signatures, and the right column are one-sentence descriptions. Headers are optional. If the table contains multiple different kinds of items (e.g. types and functions), the left column should include the kind as a suffix. [For example](https://learn.microsoft.com/en-us/dotnet/api/?view=net-8.0):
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -0,0 +1,333 @@
...
+
+> Enumerating through a collection is intrinsically not a thread-safe procedure. Even when a collection is synchronized, other threads can still modify the collection, which causes the enumerator to throw an exception. To guarantee thread safety during enumeration, you can either lock the collection during the entire enumeration or catch the exceptions resulting from changes made by other threads.
+
+#### *Usually* don't refer to "we" or "us"
+
+Use first-person pronouns sparingly to draw attention to non-technical information like design advice. Always use the first-person plural (i.e. "we" or "us") and never the singular (i.e. "I" or "me"). Often these should be accompanied by marker words like "recommend," "advise," "encourage" or "discourage." [For example](https://learn.microsoft.com/en-us/dotnet/api/system.collections.arraylist?view=net-8.0):
+
+> We don't recommend that you use the `ArrayList` class for new development. Instead, we recommend that you use the generic `List<T>` class.
+
```
cloutiertyler, line 95:
```suggestion
#### *Usually* Avoid Passive Voice

Use active voice rather than passive voice to avoid ambiguity regarding who is doing the action. Active voice directly attributes actions to the subject, making sentences easier to understand. For example:

- Passive voice: "The method was invoked."
- Active voice: "The user invoked the method."

The second example is more straightforward and clarifies who is performing the action. In most cases, prefer using the active voice to maintain a clear and direct explanation of code behavior.

However, passive voice may be appropriate in certain contexts where the actor is either unknown or irrelevant. In these cases, the emphasis is placed on the action or result rather than the subject performing it. For example:

- "The `Dispose` method is called automatically when the object is garbage collected."
```

STYLE.md
```diff
@@ -0,0 +1,333 @@
...
+#### Child items
+
+If the described item has any children (e.g. properties and methods of classes, variants of enums), include one or more tables for those children, as described above, followed by subsections for each child item. These subsections follow the same format as for the parent items, with a header, declaration, description, examples and tables of any (grand-)children.
+
+If a documentation page ends up with more than 3 layers of nested items, split it so that each top-level item has its own page.
+
+### Grammars and syntax
+
+TODO: Figure out the best way to format grammars. What's in the SQL ref looks good. Find out how we made the flow chart graphics, and encourage using those.
```
cloutiertyler, line 201: I believe Mario made them, however Josh opted to go with a text based ebnf for SQL in the future: https://github.com/clockworklabs/SpacetimeDBPrivate/edit/master/proposals/0011-sql-api.md

Which tbh I find more readable and simplier.

cloutiertyler (approved):
Yeah I mean this is excellent. I left two suggestions. I really appreciate this.

STYLE.md
gefjon: I don't love EBNF for broad-appeal reference documentation. I don't hate it either, but I think a lot of our users will look at it, not know what it is or how to read it, and give up on reading the docs. I don't particularly think that the flow charts are better by this metric, though. The EBNF at least has the advantage that it's unambiguous and familiar to a certain subset of our users. Perhaps what we want is to use the EBNF, but then to also include a whole lot of examples so that people who aren't used to thinking about CFGs, but are able to read SQL snippets.

PR #113, opened 2024-11-26
Describe how to define multiple tables with the same type

### LINKED ISSUE #90: docs: Describe how to define multiple tables with the same type
The docs should describe how to define multiple table instances from the same type.

RReverser:
Fixes #90.

### DIFF
--- a/docs/modules/c-sharp/index.md
+++ b/docs/modules/c-sharp/index.md
@@ -256,8 +256,8 @@ public partial struct Person
 You can create multiple tables backed by items of the same type by applying it with different names. For example, to store active and archived posts separately and with different privacy rules, you can declare two tables like this:
 
 ```csharp
-[SpacetimeDB.Table(Name = "posts", Public = true)]
-[SpacetimeDB.Table(Name = "archived_posts", Public = false)]
+[SpacetimeDB.Table(Name = "Posts", Public = true)]
+[SpacetimeDB.Table(Name = "ArchivedPosts", Public = false)]
 public partial struct Post {
     public string Title;
     public string Body;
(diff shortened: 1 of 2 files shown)


docs/modules/c-sharp/index.md
```diff
@@ -253,6 +253,17 @@ public partial struct Person
 }
 ```
 
+You can create multiple tables backed by items of the same type by applying it with different names. For example, to store active and archived posts separately and with different privacy rules, you can declare two tables like this:
+
+```csharp
+[SpacetimeDB.Table(Name = "Posts", Public = true)]
+[SpacetimeDB.Table(Name = "ArchivedPosts", Public = false)]
```
gefjon, line 260: Table names should be singular, not plural.
```suggestion
[SpacetimeDB.Table(Name = "Post", Public = true)]
[SpacetimeDB.Table(Name = "ArchivedPost", Public = false)]
```

docs/modules/rust/index.md
```diff
@@ -165,6 +165,17 @@ struct Person {
 }
 ```
 
+You can create multiple tables backed by items of the same type by applying it with different names. For example, to store active and archived posts separately and with different privacy rules, you can declare two tables like this:
+
+```rust
+#[table(name = posts, public)]
+#[table(name = archived_posts)]
```
gefjon, line 172:
```suggestion
#[table(name = post, public)]
#[table(name = archived_post)]
```

gefjon:
I am not sure how much value there is in incrementally editing these files, since we are going to rewrite them entirely.

docs/modules/c-sharp/index.md
RReverser: Sigh. As a user, I still disagree with it and would choose plural, but fine :P

gefjon (approved):

PR #127, opened 2025-01-07
How-to: Incremental Migrations

gefjon:
This commit adds a how-to guide for defining "incremental migrations," a strategy for updating the schema of a database while maintaining compatibility with outdated clients and without requiring a manual migration.

The code is not on GitHub yet, as I'd like review on my choice of example before pushing the repository. As such, the links to the code at the bottom of the new document are broken.


docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
+# Incremental Migrations
+
+SpacetimeDB does not contain support for general schema-modifying migrations. It does, however, allow adding new tables, and changing reducers' definitions in arbitrary ways. It's possible to run general migrations using an external tool, but this is tedious, necessitates downtime, and imposes the requirement that you update all your clients at the same time as publishing your new module version.
```
cloutiertyler, line 3:
```suggestion
SpacetimeDB does not yet provide built-in support for general schema-modifying migrations. It does, however, allow adding new tables, and changing reducers' definitions in arbitrary ways. It's possible to run general migrations using an external tool, but this is tedious, necessitates downtime, and imposes the requirement that you update all your clients at the same time as publishing your new module version.
```

docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
...
+    );
+}
+```
+
+We'll play around a bit with `spacetime call` to set up a character:
+
+```sh
+$ spacetime logs incr-migration-demo -f &
+
+$ spacetime call incr-migration-demo create_character '{ "Fighter": {} }' "Phoebe"
```
kazimuth, line 105: Do we explain this enum syntax somewhere in the docs? That seems worth doing & linking to.

docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
...
+    #[primary_key]
+    player_id: Identity,
+    nickname: String,
+    level: u32,
+    class: Class,
+    alliance: Alliance,
+}
+```
+
+When a new player creates a character, we'll make rows in both tables for them. This way, any old clients that are still subscribing to the original `character` table will continue to work, though of course they won't know about the character's alliance.
```
kazimuth, line 184: You know, we may also want a tutorial on writing ECS-style code. That paradigm fits into SQL well enough and would achieve what this guide achieves. (You still want both guides because in some cases you need this.) Unfortunately, we don't have good support for the foreign-key constraints you'd ideally want when writing ECS code.

docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
...
+        class,
+    });
+    ctx.db.character_v2().insert(CharacterV2 {
+        player_id: ctx.sender,
+        nickname,
+        level: 1,
+        class,
+        alliance: Alliance::Neutral,
+    });
+}
```
kazimuth, line 206: If the database developer is using any `auto_inc` fields in `Character`, they will need to choose between:

1. Creating *new* `auto_inc` fields in `CharacterV2`, which will be out of sync with those in `Character`.
2. Or, making the new fields in `CharacterV2` non-auto-inc, with values mirrored from `Character`.

This pattern is correct for choice 1 and incorrect for choice 2.

kazimuth: In general I'm unhappy that there's multiple places where `CharacterV2` is initialized from `Character`. In this case you keep them consistent but it's easy to screw up with more data.

I would recommend instead creating a `Character` here and calling `find_character_for_player`.

docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
...
+```
+
+We'll update our helper functions so that they operate on `character_v2` rows. In `find_character_for_player`, if we don't see the player's row in `character_v2`, we'll migrate it from `character` on the fly. In this case, we'll make the player neutral, since they haven't chosen an alliance yet.
```
kazimuth, line 209: It is critically important that the developer do this for ALL module code that operates on `character` rows.
Clients need to be able to see the old data structure, but the server should only operate on the new data structure. Server code that operates on the old data structure can result in tables getting out of sync.

A straightforward way to enforce this is using `#[deprecated]` on `Character` and only using `#[allow(deprecated)]` in `update_character` and `find_character_for_player`. The equivalents in `C#` are [`[Obsolete("dont use this", true)]`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/general#obsolete-attribute) and [`[SuppressMessageAttribute(...)]`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute?view=net-9.0).

docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
...
+    update_character(
+        ctx,
+        CharacterV2 {
+            alliance,
+            ..character
+        },
+    );
+}
+```
+
```
kazimuth, line 308: Database developers will also want to add a scheduled reducer that checks that the two tables are in sync.

docs/how-to/incremental-migrations.md
```diff
@@ -0,0 +1,365 @@
...
+ <snip>    | "Gefjon" | 3     | (Fighter = ()) | (Good = ())
+
+# But `character` is not changed, since it doesn't know about alliances:
+$ spacetime sql incr-migration-demo 'SELECT * FROM character'
+
+ player_id | nickname | level | class
+-----------+----------+-------+----------------
+ <snip>    | "Gefjon" | 3     | (Fighter = ())
+```
+
```
kazimuth, line 364: To maintain something like this long-term, we'd like recurring checks that data is synced up across versions. I would recommend a repeating reducer that runs hourly, say, that does this check.

It would be even better to write a check constraint on the join between the data tables, but we don't support that yet.

kazimuth (commented):
This looks good. I would recommend extreme caution around keeping this data in sync. If it gets out of sync it will be a problem.

docs/how-to/incremental-migrations.md
gefjon: Hm, you're right that auto_inc fields are complex. I think the correct solution is:
- Any fields that are in both tables should be auto_inc only in v1, and mirrored in v2.
- Any fields that are newly added in v2 should be auto_inc there.

> In general I'm unhappy that there's multiple places where CharacterV2 is initialized from Character. In this case you keep them consistent but it's easy to screw up with more data.

What do you mean? There's only one place where a `character` row gets translated to `character_v2` in this code. There's also one place where new rows get created, and one place where rows get updated.

docs/how-to/incremental-migrations.md
gefjon: Oh, good call.

docs/how-to/incremental-migrations.md
gefjon: This seems like an anti-pattern and also like a great way to write stupidly slow reducers that block your whole game.

docs/how-to/incremental-migrations.md
kazimuth: https://github.com/orgs/clockworklabs/projects/22/views/16?pane=issue&itemId=92869068 would be even better for this in the future.

docs/how-to/incremental-migrations.md
gefjon: I'm comfortable saying "provide built-in support" rather than "contain support," but I don't think we should say "yet" here. As I noted in the style guide, I think we need to keep our plans for future features confined to the road map, so that our docs can be concerned with the state of the world as it is today. When we use phrasings like "does not yet provide," it gives readers the sense that we're apologizing for known flaws in SpacetimeDB. We want to present SpacetimeDB 1.0 as a complete tool that's sufficient for building real-world applications already, and that means owning the choices we've made not to provide certain features.

docs/how-to/incremental-migrations.md
kazimuth: Yeah but having out-of-sync tables is a recipe for item duplication bugs and the like.

I guess you could do such a query manually with the CLI...

docs/how-to/incremental-migrations.md
kazimuth: `CharacterV2`s are initialized at both 199:205 and 225:230. It's a different flow when creating a fresh `CharacterV2` versus initializing a `CharacterV2` backwards-compatibly from a `CharacterV1`.

Really it's a case-by-case issue. It reminds me of lightfox's chains of migrations. When initializing a new row of an incrementally migrated table, do you:
- Create the newest version of the table, and then write-back to the old versions?
- Or, create the oldest version of the table, and then run your migration chain to default-initialize the new data?

If you have auto-inc fields in the old table, you need to start old and then roll your changes forward. If you don't, you can do whatever.

docs/how-to/incremental-migrations.md
kazimuth: Another solution would be:
- Write code that asserts the integrity of synced tables for a particular row ID.
- Call this after any write to synced tables, in particular after any initialization or update code.

This will be mildly slow but I think it's a small price to pay to guarantee database integrity.

kazimuth: I guess `update_character` suffices for enforcement of this. See my other comment above.

docs/how-to/incremental-migrations.md
kazimuth: The happy medium is:

- Create the oldest version of the row, run your migration chain to initialize new versions of the row, then call the write-back function on the newest version.

That is:
```
fn create_character(ctx: &ReducerCtx) {
    let v1 = character.insert(Character { /* ... */ });
    let v2 = find_character(ctx, v1.id);
    update_character(ctx, v2);
}
```

This means that the same initialization logic happens for all rows of the new table. It also means that there's a single point of enforcement for the sync logic (`update_character`), and this logic has always been run for all rows of the new table.

docs/how-to/incremental-migrations.md
gefjon: It's in the JSON SATN ref. I'll add a link.

docs/how-to/incremental-migrations.md
gefjon: Unfortunately we can't (really) do this, as applying `#[table]` to a struct marked `#[deprecated]` will result in deprecated warnings from the macro-generated code. It's possible to work around, but only in an ugly way that we don't really want to put in our docs and explain:

```rust
mod character_v1 {
    #![allow(deprecated)]

    use super::*;
    #[spacetimedb::table(name = character, public)]
    #[deprecated = "Replaced with the CharacterV2 type and the character_v2 table"]
    pub struct Character {
        #[primary_key]
        pub player_id: Identity,
        pub nickname: String,
        pub level: u32,
        pub class: Class,
    }
}
use character_v1::*;
```

docs/how-to/incremental-migrations.md
gefjon: I've removed the `auto_inc` field. I think this document as it stands now is enough to set people in the right direction, at least, and we can add more in the future.

kazimuth (approved):
This looks alright, I have no complaints.

PR #129, opened 2025-01-10
Style guide: add formatting advice for GUI elements and menu paths

gefjon:
@cloutiertyler requested that we codify in the style guide how to format paths through GUI menu items. I think the specific formatting doesn't really matter, so long as we're consistent. As such, I've chosen the first thing that came to mind: bold text for any words the reader should look for on their screen, and the thin arrow as a separator in submenu paths.

This means that the example Tyler asked me for is formatted like:

1. Open the Unity Package Manager by going to **Window -> Package Manager**.
2. In the Package Manager window, select the **Unity Registry** tab to view Unity packages [...].


kazimuth (approved):
👍

PR #133, opened 2025-01-15
Add remark to style guide inspired by SpacetimeDB#2050

kazimuth:
I started doing this unconsciously while writing SpacetimeDB#2050 and I think it's worth formalizing. Users might get very confused if they try to copy-paste our meta-syntax.


gefjon (approved):
This is an excellent idea.

PR #135, opened 2025-01-16
Made fixes to the tutorial for changes that were introduced to the C# in RC3

cloutiertyler:
I have tested this in the tutorial. Obviously it only works against RC3 APIs.


rekhoff (approved):
Simple changes with a known scope and implications. Approved!

PR #155, opened 2025-01-27
Changed subscript to SubscribeToAllTables

rekhoff:
Missed updating the HandleConnect from `Subscribe("SELECT * FROM *"` to `SubscribeToAllTables()` in the updated code block on page 3.


cloutiertyler (approved):
Thanks for updating, I think I have a ticket to do this that I didn't get to. LGTM

PR #166, opened 2025-02-14
API for mutable subscriptions

### LINKED ISSUE #78: Subscriptions: Document the new API
This includes the WebSocket API and the client SDKs as well as any example projects that use them.

joshua-spacetime:
Closes #78.

Includes rust and csharp examples.

### DIFF
--- a/docs/subscriptions/index.md
+++ b/docs/subscriptions/index.md
@@ -1,0 +1,446 @@
+# The SpacetimeDB Subscription API
+
+The subscription API allows a client to replicate a subset of a database.
+It does so by registering SQL queries, which we call subscriptions, through a database connection.
+A client will only receive updates for rows that match the subscriptions it has registered.
+
+For more information on syntax and requirements see the [SQL docs](/docs/sql#subscriptions).
+
+This guide describes the two main interfaces that comprise the API - `SubscriptionBuilder` and `SubscriptionHandle`.
+By using these interfaces, you can create efficient and responsive client applications that only receive the data they need.
+
+## SubscriptionBuilder
+
+:::server-rust
+```rust
+pub struct SubscriptionBuilder<M: SpacetimeModule> { /* private fields */ }
+
+impl<M: SpacetimeModule> SubscriptionBuilder<M> {
+    /// Register a callback that runs when the subscription has been applied.
+    /// This callback receives a context containing the current state of the subscription.
+    pub fn on_applied(mut self, callback: impl FnOnce(&M::SubscriptionEventContext) + Send + 'static);
+
+    /// Register a callback to run when the subscription fails.
+    ///
+    /// Note that this callback may run either when attempting to apply the subscription,
+    /// in which case [`Self::on_applied`] will never run,
+    /// or later during the subscription's lifetime if the module's interface changes,
+    /// in which case [`Self::on_applied`] may have already run.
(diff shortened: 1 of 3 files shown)


docs/subscriptions/index.md
```diff
@@ -0,0 +1,273 @@
...
+var conn = ConnectToDB();
+
+// Register a subscription with the database
+var userSubscription = conn
+    .SubscriptionBuilder()
+    .OnApplied((ctx) => { /* handle applied state */ })
+    .OnError((errorCtx, error) => { /* handle error */ })
+    .Subscribe("SELECT * FROM user");
+```
+:::
```
gefjon, line 139: I feel like the two langs should use the same query. The code should be as equivalent as possible.

docs/subscriptions/index.md
```diff
@@ -0,0 +1,273 @@
...
+Once applied, a client will start receiving row updates to its client cache.
+A client can react to these updates by registering row callbacks for the appropriate table.
+
+It is important to note that subscriptions must be disjoint from one another.
+Two subscriptions that return the same row may result in a corrupted client cache.
+That is, one that does not accurately and consistently reflect the state of the database.
+
+### Example Usage
+
+:::server-rust
```
gefjon, line 114: Do these have to be  `server-rust` and `server-csharp`? It's actually client code we're showing. CC @cloutiertyler , as I'm not actually sure how the code toggle thing works.

gefjon (commented):
Page needs to describe what queries are allowed as subscriptions - likely just by linking to the SQL doc you already wrote. It should also call out that overlapping queries aren't allowed within this document, as that seems like it'll get a lot of people and we don't have solid error reporting for it. And again, I'd prefer not to say "yet" about overlapping queries when you do add that section.

docs/subscriptions/index.md
joshua-spacetime: From conversation with Tyler, only `server-rust` and `server-csharp` are supported right now. It used to be just `:::rust` and `:::csharp`, but the idea was that you might want to specify your client and server language separately, although this is yet to be implemented.

docs/subscriptions/index.md
```diff
@@ -0,0 +1,446 @@
...
+```
+
+All other subscriptions continue to remain in effect.
+:::
+
+## Best Practices for Optimizing Server Compute and Reducing Serialization Overhead
+
+### 1. Writing Efficient SQL Queries
+
+For writing efficient SQL queries, see our [SQL Best Practices guide](/docs/sql).
```
joshua-spacetime, line 286: This should link to the "Best Practices" subsection within the sql docs.

docs/subscriptions/index.md
gefjon: How does it render? If it actually shows the word "server," I think that's a deal-breaker. If not, then we're fine, though it would very much be preferable to be able to also have a TypeScript option.

gefjon (approved):

docs/subscriptions/index.md
joshua-spacetime: It will render as "Module Language", which could be confusing.

gefjon: Ugh. Hmm. Let's talk about this in the meeting.

joshua-spacetime: I spoke with @cloutiertyler. He's going to remove the "Module" distinction. See issue [tracker](https://github.com/clockworklabs/spacetime-web/issues/513).

PR #168, opened 2025-02-18
CLI docs

bfops:
Add a CLI reference section.

It looks roughly like this:
image

I think we could do a lot of styling improvements here, but I think this is at least "acceptable", so I think we should merge this and iterate from there.

## Note:
This depends on https://github.com/clockworklabs/SpacetimeDB/pull/2276 in order to generate it. Instructions are in the README.

I had to do some manual tweaking of the markdown because our website doesn't seem to quite parse it properly in headers. I've documented those steps in the README as well.

# Testing
If you want to check it out yourself, point your `spacetime-web` repo at `bfops/cli-docs` and go to `http://localhost:5173/docs/cli-reference`

- [x] The `spacetime publish` link at the top of the page does successfully scroll to the right section

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -34,7 +34,7 @@ This document contains the help content for the `spacetime` command-line program
 * [`spacetime start`↴](#spacetime-start)
 * [`spacetime version`↴](#spacetime-version)
 
-## <code>spacetime</code>
+## spacetime
 
 **Usage:** `spacetime [OPTIONS] <COMMAND>`
 
@@ -82,7 +82,7 @@ WARNING: This command is UNSTABLE and subject to breaking changes.
 
 
 
-## <code>spacetime publish</code>
+## spacetime publish
 
 Create and update a SpacetimeDB database
 
@@ -110,7 +110,7 @@ Run `spacetime help publish` for more detailed information.
 
 
 
-## <code>spacetime delete</code>
+## spacetime delete
 
 Deletes a SpacetimeDB database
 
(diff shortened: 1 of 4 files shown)


docs/cli-reference.md
```diff
@@ -0,0 +1,589 @@
...
+**Usage:** `spacetime logs [OPTIONS] <database>`
+
+Run `spacetime help logs` for more detailed information.
+
+
+###### <b>Arguments:</b>
+
+* `<DATABASE>` — The name or identity of the database to print logs from
+
+###### <b>Options:</b>
```
cloutiertyler, line 130: The h6 + <b> tag is an odd choice, I assume this is a styling workaround?

bfops: "Yes", but what's autogenerated is `###### **Options:**`.

It does look a bit better when it's bolded, so I switched `**` for a `<b>` tag. I can change it for e.g. `<h6><b>`, but then I'd want to re-test that it looks as expected.

cloutiertyler: `**` is bold though.

I just mean why not just bold it? Why the tiny header?

bfops: I don't know, but that's what's generated. I just swapped the markdown bold (`**`) that didn't work for an HTML bold (`<b>`) that did.

We can do further post-processing to remove that if you feel strongly about it. My hope was to remove the hacky `sed`s if/when the generated stuff just starts working.

cloutiertyler: Alright, I'm willing to fix this/change this in a future PR. This is an improvement in my mind.

cloutiertyler (approved):

PR #172, opened 2025-02-20
Update Rust SDK ref, and also a few small fixes

gefjon:
Mostly updates to the Rust SDK ref. Also has a couple tiny commits because I thought this would be a PR with a bunch of  tiny fixes, but then updating the SDK ref became big.


docs/sdks/rust/index.md
```diff
@@ -177,29 +334,114 @@ trait DbContext {
...
+##### Method `subscribe`
 
 ```rust
 impl SubscriptionBuilder {
     fn subscribe(self, queries: impl IntoQueries) -> SubscriptionHandle;
 }
 ```
 
-Subscribe to a set of queries. `queries` should be an array or slice of strings.
+Subscribe to a set of queries. `queries` should be a string or an array, vec or slice of strings.
```
kazimuth, line 366: link to sql docs

gefjon: Added in a paragraph directly after this one.

docs/sdks/rust/index.md
```diff
@@ -177,29 +331,116 @@ trait DbContext {
...
+
+##### Method `subscribe_to_all_tables`
+
+```rust
+impl SubscriptionBuilder {
+    fn subscribe_to_all_tables(self);
+}
+```
+
+Subscribe to all rows from all tables. This method is provided as a convenience for simple clients. The same connection must not mix `subscribe_to_all_tables` with [`subscribe` to specific queries](#method-subscribe). Doing so may cause errors or corrupt the client cache, leading to local queries returning incorrect results. The subscription initiated by `subscribe_to_all_tables` cannot be canceled after it is initiated.
```
joshua-spacetime, line 375:
```suggestion
Subscribe to all rows from all public tables. This method is provided as a convenience for simple clients. The subscription initiated by `subscribe_to_all_tables` cannot be canceled after it is initiated. You should [`subscribe` to specific queries](#method-subscribe) if you need fine-grained control over the lifecycle of your subscriptions.
```

Technically mixing both calls will no longer cause errors or corrupt the cache, but it still doesn't make sense to do it.

docs/sdks/rust/index.md
```diff
@@ -349,12 +669,135 @@ module_bindings::Reducer
...
 
+| Name                                                              | Description                                                                     |
+|-------------------------------------------------------------------|---------------------------------------------------------------------------------|
+| [`Table` trait](#trait-table)                                     | Provides access to subscribed rows of a specific table within the client cache. |
+| [`TableWithPrimaryKey` trait](#trait-tablewithprimarykey)         | Extension trait for tables which have a column designated as a primary key.     |
+| [Unique constraint index access](#unique-constraint-index-access) | Seek a subscribed row by the value in its unique or primary key column.         |
+| [BTree index access](#btree-index-access)                         | Not supported.                                                                  |
```
joshua-spacetime, line 799: Any reason to call this out as not supported vs just not mentioning it?

docs/sdks/rust/index.md
```diff
@@ -451,17 +906,17 @@ For each unique constraint on a table, its table handle has a method whose name
 
 ### BTree index access
 
-Not currently implemented in the Rust SDK. Coming soon!
+The SpacetimeDB Rust client SDK does not support non-unique BTree indexes.
```
joshua-spacetime, line 909: Same here. Does this need to be called out explicitly?

docs/sdks/rust/index.md
```diff
@@ -81,17 +111,31 @@ This interface may change in an upcoming release as we rework SpacetimeDB's auth
...
+    fn on_connect_error(
+        self,
+        callback: impl FnOnce(&ErrorContext, spacetimedb_sdk::Error),
+    ) -> DbConnectionBuilder;
+}
+```
+
+Chain a call to `.on_connect_error(callback)` to your builder to register a callback to run when your connection fails.
+
+A known bug in the SpacetimeDB Rust client SDK currently causes this callback never to be invoked. [`on_disconnect`](#callback-on_disconnect) callbacks are invoked instead.
```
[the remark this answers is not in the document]
gefjon: I prefer to get out in front of things like this. I could be convinced otherwise, but my expectation is that if we don't have this here, we're going to get questions about why `on_connect_error` isn't getting called, even if we never mention it on the website, since autocomplete and rustdoc will show it.

docs/sdks/rust/index.md
```diff
@@ -129,45 +178,150 @@ impl DbConnection {
...
+spacetimedb_sdk::SubscriptionBuilder
+```
+
+| Name                                                                             | Description                                                     |
+|----------------------------------------------------------------------------------|-----------------------------------------------------------------|
+| [`ctx.subscription_builder()` constructor](#constructor-ctxsubscription_builder) | Begin configuring a new subscription.                           |
+| [`on_applied` callback](#callback-on_applied)                                    | Register a callback to run when matching rows become available. |
+| [`on_error` callback](#callback-on_error)                                        | Register a callback to run if the subscription fails.           |
+| [`subscribe` method](#method-subscribe)                                          | Finish configuration and subscribe to one or more SQL queries.  |
+| [`subscribe_to_all_tables` method](#method-subscribe_to_all_tables)              | Convenience method to subscribe to the entire database.         |
```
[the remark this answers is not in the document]
gefjon: Yes, that's a good idea.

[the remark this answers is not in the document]
gefjon: Well, hm. No, I think this blurb should be kept short, and we should keep the more in-depth discussion in the method's actual docs.

docs/sdks/rust/index.md
gefjon: Again, I could be convinced to remove this, but I stuck it in because a lot of other contexts, incl. modules and the C# client SDK, *do* support non-unique BTrees, and I wanted to reduce confusion due to the asymmetry.

joshua-spacetime (approved):

PR #178, opened 2025-02-24
Document LIMIT and COUNT

### LINKED ISSUE #177: Update sql docs to include LIMIT and COUNT

joshua-spacetime:
Closes #177.

### DIFF
--- a/docs/sql/index.md
+++ b/docs/sql/index.md
@@ -157,14 +157,15 @@ SELECT * FROM Inventory WHERE price > {X} AND amount < {Y}
 ### SELECT
 
 ```ebnf
-SELECT projection FROM relation [ WHERE predicate ]
+SELECT projection FROM relation [ WHERE predicate ] [LIMIT NUM]
 ```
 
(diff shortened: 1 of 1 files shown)


docs/sql/index.md
```diff
@@ -219,6 +228,19 @@ WHERE product.name = {product_name}
 
 See [Subscriptions](#where).
 
+#### LIMIT clause
+
+Limits the number of rows returned by a query.
+
+It does not order rows in any way.
```
mamcx, line 235:
```suggestion
No more than that many rows will be returned (but possibly fewer, if the query itself yields fewer rows)
```

I don't think is clarifying to talk about ordering

mamcx (commented):
I think we need to add a section for `count`:


```midtown
## Aggregate Functions

Aggregate function computes a single result from multiple input row.

### `COUNT(*)`

Counts the number of rows returned
```

mamcx (approved):
LGTM

PR #181, opened 2025-02-24
Update TypeScript SDK reference

gefjon:
Based on #172 . What I did was, I deleted the whole existing TypeScript ref, pasted in the new Rust ref, and then replaced all the Rust code with TypeScript.

I'm not super happy with how I'm formatting the code blocks, but :shrug: I think it's good enough.

### DIFF
--- a/docs/sdks/typescript/index.md
+++ b/docs/sdks/typescript/index.md
@@ -2,9 +2,21 @@
 
 The SpacetimeDB client SDK for TypeScript contains all the tools you need to build clients for SpacetimeDB modules using Typescript, either in the browser or with NodeJS.
 
-> You need a database created before use the client, so make sure to follow the Rust or C# Module Quickstart guides if need one.
+| Name                                                              | Description                                                                                                                            |
+|-------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| [Project setup](#project-setup)                                   | Configure a Rust crate to use the SpacetimeDB Rust client SDK.                                                                         |
+| [Generate module bindings](#generate-module-bindings)             | Use the SpacetimeDB CLI to generate module-specific types and interfaces.                                                              |
+| [`DbConnection` type](#type-dbconnection)                         | A connection to a remote database.                                                                                                     |
+| [`DbContext` trait](#trait-dbcontext)                             | Methods for interacting with the remote database. Implemented by [`DbConnection`](#type-dbconnection) and various event context types. |
+| [`EventContext` type](#type-eventcontext)                         | [`DbContext`](#trait-dbcontext) available in [row callbacks](#callback-on_insert).                                                     |
+| [`ReducerEventContext` type](#type-reducereventcontext)           | [`DbContext`](#trait-dbcontext) available in [reducer callbacks](#observe-and-invoke-reducers).                                        |
+| [`SubscriptionEventContext` type](#type-subscriptioneventcontext) | [`DbContext`](#trait-dbcontext) available in [subscription-related callbacks](#subscribe-to-queries).                                  |
+| [`ErrorContext` type](#type-errorcontext)                         | [`DbContext`](#trait-dbcontext) available in error-related callbacks.                                                                  |
+| [Access the client cache](#access-the-client-cache)               | Make local queries against subscribed rows, and register [row callbacks](#callback-on_insert) to run when subscribed rows change.      |
+| [Observe and invoke reducers](#observe-and-invoke-reducers)       | Send requests to the database to run reducers, and register callbacks to run when notified of reducers.      |
+| [Identify a client](#identify-a-client)                           | Types for identifying users and client connections.                                                                                    |
 
-## Install the SDK
+## Project setup
 
 First, create a new client project, and add the following to your `tsconfig.json` file:
 
@@ -55,23 +67,7 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
 mkdir -p client/src/module_bindings
 spacetime generate --lang typescript \
     --out-dir client/src/module_bindings \
-    --project-path server
-```
-
-And now you will get the files for the `reducers` & `tables`:
-
-```bash
-quickstart-chat
-├── client
-│   ├── node_modules
-│   ├── public
-│   └── src
-|       └── module_bindings
-|           ├── add_reducer.ts
-|           ├── person.ts
-|           └── say_hello_reducer.ts
-└── server
-    └── src
+    --project-path PATH-TO-MODULE-DIRECTORY
 ```
 
 Import the `module_bindings` in your client's _main_ file:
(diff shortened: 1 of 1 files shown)


docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
-| ----------------------------------------------- | -------------------------------------------- |
-| [`Identity.constructor`](#identity-constructor) | Creates a new `Identity`.                    |
-| Methods                                         |                                              |
-| [`Identity.isEqual`](#identity-isequal)         | Compare two identities for equality.         |
-| [`Identity.toHexString`](#identity-tohexstring) | Print the identity as a hexadecimal string.  |
-| Static methods                                  |                                              |
-| [`Identity.fromString`](#identity-fromstring)   | Parse an Identity from a hexadecimal string. |
+Subscribe to all rows from all tables. This method is provided as a convenience for simple clients. The same connection must not mix `subscribeToAllTables` with [`subscribe` to specific queries](#method-subscribe). Doing so may cause errors or corrupt the client cache, leading to local queries returning incorrect results. The subscription initiated by `subscribeToAllTables` cannot be canceled after it is initiated.
```
joshua-spacetime, line 311:
```suggestion
Subscribe to all rows from all public tables. This method is provided as a convenience for simple clients. The subscription initiated by `subscribeToAllTables ` cannot be canceled after it is initiated. You should [`subscribe` to specific queries](#method-subscribe) if you need fine-grained control over the lifecycle of your subscriptions.
```

Same change from the rust sdk docs.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-#### Example
+##### Method `subscribeToAllTables`
 
-```ts
-spacetimeDBClient.onError((...args: any[]) => {
-  console.error('ERROR', args);
-});
+```typescript
+SubscriptionBuilder.subscribe_to_all_tables(): void
```
joshua-spacetime, line 308:
```suggestion
SubscriptionBuilder.subscribeToAllTables(): void
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
-    └── src
+    --project-path PATH-TO-MODULE-DIRECTORY
 ```
 
 Import the `module_bindings` in your client's _main_ file:
 
 ```typescript
 import { SpacetimeDBClient, Identity } from '@clockworklabs/spacetimedb-sdk';
 
-import Person from './module_bindings/person';
```
cloutiertyler, line 82: NIT: Sometimes you do this kind of import, but it's not that typical, and sometimes discouraged. More typical to do: 

```ts
import { DbConnection } from './module_bindings/index'
```

But if we do an import the variable we bind it to should follow TypeScript casing rules, so: `moduleBindings` or just `module`.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
-| `host`            | `string`               | The host of the SpacetimeDB server.                                                                                                               |
-| `name_or_address` | `string`               | The name or address of the SpacetimeDB module.                                                                                                    |
-| `auth_token?`     | `string`               | The credentials to use to connect to authenticate with SpacetimeDB.                                                                               |
-| `protocol?`       | `"binary"` \| `"json"` | Define how encode the messages: `"binary"` \| `"json"`. Binary is more efficient and compact, but JSON provides human-readable debug information. |
+Construct a `DbConnection` by calling `DbConnection.builder()` and chaining configuration methods, then calling `.build()`. You must at least specify `withUri`, to supply the URI of the SpacetimeDB to which you published your module, and `with_module_name`, to supply the human-readable SpacetimeDB domain name or the raw `Identity` which identifies the module.
```
cloutiertyler, line 101:
```suggestion
Construct a `DbConnection` by calling `DbConnection.builder()` and chaining configuration methods, then calling `.build()`. You must at least specify `withUri`, to supply the URI of the SpacetimeDB to which you published your module, and `withModuleName`, to supply the human-readable SpacetimeDB domain name or the raw `Identity` which identifies the module.
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
-| [`SpacetimeDBClient.disconnect`](#spacetimedbclient-disconnect)   | Close the current connection.                                            |
-| [`SpacetimeDBClient.subscribe`](#spacetimedbclient-subscribe)     | Subscribe to a set of queries.                                           |
-| Events                                                            |                                                                          |
-| [`SpacetimeDBClient.onConnect`](#spacetimedbclient-onconnect)     | Register a callback to be invoked upon authentication with the database. |
-| [`SpacetimeDBClient.onError`](#spacetimedbclient-onerror)         | Register a callback to be invoked upon a error.                          |
+A connection to a remote database is represented by the `module_bindings.DbConnection` type. This type is generated per-module, and contains information about the types, tables and reducers defined by your module.
 
-## Constructors
+| Name                                                      | Description                                                                                      |
```
cloutiertyler, line 89: This is a table of contents, right? Maybe we should call this "Section" instead of "Name"?

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-### `SpacetimeDBClient` constructor
 
-Creates a new `SpacetimeDBClient` database client and set the initial parameters.
+### Connect to a module
 
-```ts
-new SpacetimeDBClient(host: string, name_or_address: string, auth_token?: string, protocol?: "binary" | "json")
+```typescript
+DbConnection.builder(): DbConnectionBuilder
```
cloutiertyler, line 98:
```suggestion
builder(): DbConnectionBuilder
```

Or

```ts
class DbConnectionBuilder {
  public withUri(uri: string): DbConnectionBuilder;
}

If you prefer. The above implies that it is a class method.

Comment (cloutiertyler) [reply docs/sdks/typescript/index.md:98]: Let's do the latter I guess.

```suggestion
class DbConnection {
  public builder(): DbConnectionBuilder;
}
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
+#### Method `withUri`
 
-var spacetimeDBClient = new SpacetimeDBClient(
-  host,
-  name_or_address,
-  auth_token,
-  protocol
-);
+```typescript
+DbConnectionBuilder.withUri(uri: string): DbConnectionBuilder
```
cloutiertyler, line 116:
```suggestion
class DbConnectionBuilder {
  public withUri(uri: string): DbConnectionBuilder
}
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-## Class methods
+Configure the URI of the SpacetimeDB instance or cluster which hosts the remote module.
 
-### `SpacetimeDBClient.registerReducers`
+#### Method `withModuleName`
 
-Registers reducer classes for use with a SpacetimeDBClient
+```typescript
+DbConnectionBuilder.withModuleName(name_or_identity: string): DbConnectionBuilder
```
cloutiertyler, line 124: Probably easier to update this in a proper editor rather than adding suggestions, but ditto on all of them.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-```ts
-registerReducers(...reducerClasses: ReducerClass[])
 ```
 
-#### Parameters
+Configure the SpacetimeDB domain name or `Identity` of the remote module which identifies it within the SpacetimeDB instance or cluster.
```
cloutiertyler, line 128:
```suggestion
Configure the SpacetimeDB domain name or hex string encoded `Identity` of the remote module which identifies it within the SpacetimeDB instance or cluster.
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
+DbConnectionBuilder.onConnect(
+    callback: (ctx: DbConnection, identity: Identity, token: string) => void
+): DbConnectionBuilder
 ```
 
----
+Chain a call to `.onConnect(callback)` to your builder to register a callback to run when your new `DbConnection` successfully initiates its connection to the remote module. The callback accepts three arguments: a reference to the `DbConnection`, the `Identity` by which SpacetimeDB identifies this connection, and a private access token which can be saved and later passed to [`withToken`](#method-withtoken) to authenticate the same user in future connections.
 
-### `SpacetimeDBClient.registerTables`
+This interface may change in an upcoming release as we rework SpacetimeDB's authentication model.
```
cloutiertyler, line 140: Will it? I was not aware of planned changes. But also I don't think we should say this, because this either saying it's not stable (even though we're claiming 1.0) or it's coming in 2.0 in which case we should just not say it.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-### `SpacetimeDBClient` identity
+#### Method `withToken`
 
-The user's public [Identity](#class-identity).
-
-```
-identity: Identity | undefined
+```typescript
+DbConnectionBuilder.withToken(token: string | null): DbConnectionBuilder
```
cloutiertyler, line 166:
```suggestion
DbConnectionBuilder.withToken(token: string | undefined): DbConnectionBuilder
```

JavaScript is a silly language. This is a meaningful distinction.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
-### `SpacetimeDBClient` connect
+The `reducers` field of the `DbConnection` provides access to reducers exposed by the remote module. See [Observe and invoke reducers](#observe-and-invoke-reducers).
 
-Connect to The SpacetimeDB Websocket For Your Module. By default, this will use a secure websocket connection. The parameters are optional, and if not provided, will use the values provided on construction of the client.
+## Interface `DbContext`
 
-```ts
-connect(host: string?, name_or_address: string?, auth_token: string?): Promise<void>
+```rust
+interface spacetimedb_sdk.DbContext<
```
cloutiertyler, line 201:
```suggestion
interface DbContext<
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-> A new call to `subscribe` will remove all previous subscriptions and replace them with the new `queries`.
-> If any rows matched the previous subscribed queries but do not match the new queries,
-> those rows will be removed from the client cache, and [`{Table}.on_delete`](#table-ondelete) callbacks will be invoked for them.
+#### Type `SubscriptionBuilder`
 
-```ts
-subscribe(queryOrQueries: string | string[]): void
+```typescript
+spacetimedb_sdk::SubscriptionBuilder
```
cloutiertyler, line 253: This type is actually in the module bindings, not the sdk IIRC.

docs/sdks/typescript/index.md
gefjon: I see what you mean. This one *is* what I would call a static method, or an associated function, which takes the class as a receiver. But I used this same syntax for what I would call methods, which take an instance as a receiver. What syntax would you use for each of these things, in general?

docs/sdks/typescript/index.md
gefjon: I checked the SDK code when I wrote this, and was pretty sure it said `undefined`, not `null`. But now I'm looking again, and it seems to have been changed to (or always was) `token?: string`.

That said, based on [the TypeScript handbook](https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters), it looks like `token?: string` means `string | undefined`, *not* `string | null`.

Am I missing something here?

docs/sdks/typescript/index.md
gefjon: Honestly, we could just remove the table header. But I'd prefer to do that later; for now, we're using the `| Name | Description |` formatting in several places, and it's what's listed in the style guide.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-## Constructors
+#### Type `SubscriptionHandle`
 
-### `Identity` constructor
-
-```ts
-new Identity(data: Uint8Array)
+```rust
+module_bindings::SubscriptionHandle
```
cloutiertyler, line 316: There is technically no such type in the TypeScript SDK. We just return: `SubscriptionHandleImpl<DBView, Reducers, SetReducerFlags>`

cloutiertyler: Not a huge deal though since it is that type in spirit.

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-#### Returns
+Throws an error if the subscription has already ended, either due to a previous call to `unsubscribe` or [`unsubscribeThen`](#method-unsubscribethen), or due to an error.
 
-[`Identity`](#class-identity)
+##### Method `unsubscribeThen`
 
-### Class `Address`
+```typescript
+SubscriptionHandle.unsubscribe_then(
```
cloutiertyler, line 361:
```suggestion
SubscriptionHandle.unsubscribeThen(
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-| Constructors                                  | Description                                 |
-| --------------------------------------------- | ------------------------------------------- |
-| [`Address.constructor`](#address-constructor) | Creates a new `Address`.                    |
-| Methods                                       |                                             |
-| [`Address.isEqual`](#address-isequal)         | Compare two identities for equality.        |
-| [`Address.toHexString`](#address-tohexstring) | Print the address as a hexadecimal string.  |
-| Static methods                                |                                             |
-| [`Address.fromString`](#address-fromstring)   | Parse an Address from a hexadecimal string. |
+Returns an error if the subscription has already ended, either due to a previous call to [`unsubscribe`](#method-unsubscribe) or `unsubscribe_then`, or due to an error.
```
cloutiertyler, line 369:
```suggestion
Returns an error if the subscription has already ended, either due to a previous call to [`unsubscribe`](#method-unsubscribe) or `unsubscribeThen`, or due to an error.
```

docs/sdks/typescript/index.md
```diff
@@ -55,927 +67,740 @@ Each SpacetimeDB client depends on some bindings specific to your module. Create
...
 
-| Static methods                  | Description                                                  |
-| ------------------------------- | ------------------------------------------------------------ |
-| [`Reducer.call`](#reducer-call) | Executes the reducer.                                        |
-| Events                          |                                                              |
-| [`Reducer.on`](#reducer-on)     | Register a callback to run each time the reducer is invoked. |
+### Unique constraint index access
 
-## Static methods
+For each unique constraint on a table, its table handle has a field whose name is the unique column name. This field is a unique index handle. The unique index handle has a method `.find(desired_val: Col) -> Row | undefined`, where `Col` is the type of the column, and `Row` the type of rows. If a row with `desired_val` in the unique column is resident in the client cache, `.find` returns it.
```
cloutiertyler, line 774:
```suggestion
For each unique constraint on a table, its table handle has a field whose name is the unique column name. This field is a unique index handle. The unique index handle has a method `.find(desiredValue: Col) -> Row | undefined`, where `Col` is the type of the column, and `Row` the type of rows. If a row with `desiredValue` in the unique column is resident in the client cache, `.find` returns it.
```

docs/sdks/typescript/index.md
cloutiertyler: Okay. I'll just resolve this comment.

docs/sdks/typescript/index.md
cloutiertyler: Oh, whoops, you're right. It's fine for this particular one, but confusing for the others. For static what you have here is fine:

```ts
DbConnection.builder(): DbConnectionBuilder
```

For methods let's do this:

```ts
class DbConnectionBuilder {
  public withUri(uri: string): DbConnectionBuilder
}
```

docs/sdks/typescript/index.md
cloutiertyler: I also checked. `token?: string` does mean `string | undefined`. In your docs it's `string | null` which is not correct.

docs/sdks/typescript/index.md
cloutiertyler: Ah okay, that makes more sense.

docs/sdks/typescript/index.md
gefjon: Ah, I see. Will fix. I'm going to just write `token?: string`.

docs/sdks/typescript/index.md
gefjon: If we're going to use the `class` syntax, let's write static methods as static methods, [since JS apparently supports that](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static).

docs/sdks/typescript/index.md
gefjon: Does this mean that a user who wants to store a subscription handle in some sort of data structure has to either write out `SubscriptionHandleImpl<DBView, Reducers, SetReducerFlags>` or cast to `any`?

docs/sdks/typescript/index.md
gefjon: > `import { DbConnection } from './module_bindings/index'`

We don't want to do the named import style of `import { stuff... } from path` because:
- We have to import a whole lot of stuff from the module bindings.
- In this document, we don't have a specific module that we're writing against, so we *can't* list all of the items that we'd need to import. E.g. what would the list of row types be?

> But if we do an import the variable we bind it to should follow TypeScript casing rules, so: `moduleBindings` or just `module`.

It seems weird to me to have the file name not match the imported variable name, and we've pretty firmly committed to using `module_bindings` as the directory name in all our TypeScript examples. But sure, if you think this is better.

cloutiertyler (approved):
This now looks good to me! If there are any adjustments we can make them in separate PRs. It's a definite improvement and obviously not a regression.

Issue #25, opened 2024-01-01
Document (on the website) reducer semantics with an eye to transactionality and repeating reducers


gefjon:
Such a document should cover:

- The 1:1 correspondence between host-to-WASM calls and transactions.
- The fact that a reducer which calls another reducer will not start a sub-TX.
- The fact that a `repeat` attribute on a reducer (mostly) behaves like putting a `schedule!(delay, self(ctx.timestamp));` call at the end of the reducer.
  - Different when called by another reducer? I think yes.
  - Different when the reducer fails? I think no.

bfops:
This should go in the Overview page. (If it becomes too large, we can create an "Understanding SpacetimeDB" section/page).

We should link to this section from the module reference pages.

Inspiration: https://docs.convex.dev/understanding/ and https://docs.convex.dev/functions/mutation-functions#transactions.

PR #185, opened 2025-02-26
Document reducer semantics wrt. transactionality

Centril:
Fixes https://github.com/clockworklabs/spacetime-docs/issues/25.

Not sure if parts should be pulled out to its own document, but I figure we can reorganize later?


gefjon (approved):
Looks great!

PR #188, opened 2025-02-27
Update docs for http api

coolreader18:
Decided not to add docs for `PUT,POST /energy/:identity` since it seems like those were already left out before and idk if that's really something people would be able use for maincloud.


docs/http/index.md
```diff
@@ -6,37 +6,20 @@ Rather than a password, each Spacetime identity is associated with a private tok
...
-Clients can request a new identity and token via [the `/identity POST` HTTP endpoint](/docs/http/identity#identity-post).
+Clients can request a new identity and token via [the `POST /v1/identity` HTTP endpoint](/docs/http/identity#post-v1identity).
 
 Alternately, a new identity and token will be generated during an anonymous connection via the WebSocket API, and passed to the client as an `IdentityToken` message.
 
-### Encoding `Authorization` headers
+### `Authorization` headers
 
-Many SpacetimeDB HTTP endpoints either require or optionally accept a token in the `Authorization` header. SpacetimeDB authorization headers use `Basic` authorization with the username `token` and the token as the password. Because Spacetime tokens are not passwords, and SpacetimeDB Cloud uses TLS, usual security concerns about HTTP `Basic` authorization do not apply.
+Many SpacetimeDB HTTP endpoints either require or optionally accept a token in the `Authorization` header. SpacetimeDB authorization headers are of the form `Authorization: Bearer ${token}`.
```
gefjon, line 15:
```suggestion
Many SpacetimeDB HTTP endpoints either require or optionally accept a token in the `Authorization` header. SpacetimeDB authorization headers are of the form `Authorization: Bearer ${token}`, where `token` is an [OpenID Connect](https://openid.net/developers/how-connect-works/) compliant [JSON Web Token](https://jwt.io/), such as the one returned from [the `POST /v1/identity` HTTP endpoint](/docs/http/identity#post-v1identity).
```

This is repetitive with the above section, but I think that's ok.

docs/http/index.md
```diff
@@ -6,37 +6,20 @@ Rather than a password, each Spacetime identity is associated with a private tok
 
 ### Generating identities and tokens
 
-Clients can request a new identity and token via [the `/identity POST` HTTP endpoint](/docs/http/identity#identity-post).
+Clients can request a new identity and token via [the `POST /v1/identity` HTTP endpoint](/docs/http/identity#post-v1identity).
```
gefjon, line 9:
```suggestion
SpacetimeDB can derive an identity from the `sub` and `iss` claims of any [OpenID Connect](https://openid.net/developers/how-connect-works/) compliant [JSON Web Token](https://jwt.io/).

Clients can request a new identity and token signed by the SpacetimeDB host via [the `POST /v1/identity` HTTP endpoint](/docs/http/identity#post-v1identity). Such a token will not be portable to other SpacetimeDB clusters.
```

docs/http/identity.md
```diff
@@ -133,7 +105,7 @@ Returns JSON in the form:
 
 The `addresses` value is an array of zero or more strings, each of which is the address of a database owned by the identity passed as a parameter.
 
-## `/identity/:identity/verify GET`
+## `GET /v1/identity/:identity/verify`
```
gefjon, line 108: In hindsight, we should've made this route instead just be `GET /v1/identity/verify`, take the auth header, and return the `Identity` that it auths as. In the new OIDC-based auth scheme, it's not trivial to derive the `Identity` from the token, and users may prefer to send requests to SpacetimeDB to do so rather than doing it themselves. Then again, maybe we'd design a different interface, like one where you just send `sub`/`iss` pairs without needing a signed token, or something. I don't know, we'll probably have to do some design work around this. No change required to this PR for this comment.

gefjon (commented):
I'm trusting you that these docs are correct (they look right to me, but I haven't checked). They read well and are properly formatted. I will approve as soon as we resolve the energy-related stuff (see below).

> Decided not to add docs for PUT,POST /energy/:identity since it seems like those were already left out before and idk if that's really something people would be able use for maincloud.

I agree with this choice. We'll need to have a quick out-of-band discussion to determine whether any of the `/v1/energy` routes are meaningful anymore; I suspect we may want to just rip that page out in its entirety.

gefjon:
Update: out-of-band discussion has yielded that the `/v1/energy` endpoints are now useless. As such, we should just wholesale delete that page of our docs.

gefjon (approved):

PR #191, opened 2025-02-28
C# sdk reference

joshua-spacetime:
Note to reviewers: I tried to remove any references to rust concepts, but some probably slipped through.

~~I also added a new doc instead of updating the old one to aid in review. Will replace the old doc before merging.~~

### DIFF
--- a/docs/sdks/c-sharp/index.md
+++ b/docs/sdks/c-sharp/index.md
@@ -1,56 +1,22 @@
 # The SpacetimeDB C# client SDK
 
-The SpacetimeDB client C# for Rust contains all the tools you need to build native clients for SpacetimeDB modules using C#.
+The SpacetimeDB client for C# contains all the tools you need to build native clients for SpacetimeDB modules using C#.
 
-## Table of Contents
+| Name                                                    | Description                                                               |
+|---------------------------------------------------------|---------------------------------------------------------------------------|
+| [Project setup](#project-setup)                         | Configure a C# project to use the SpacetimeDB C# client SDK.              |
+| [Generate module bindings](#generate-module-bindings)   | Use the SpacetimeDB CLI to generate module-specific types and interfaces. |
+| [`DbConnection` type](#type-dbconnection)               | A connection to a remote database.                                        |
+| [`IDbContext` interface](#interface-idbcontext)         | Methods for interacting with the remote database.                         |
+| [`EventContext` type](#type-eventcontext)               | Implements [`IDbContext`](##interface-idbcontext) for [row callbacks](#callback-oninsert). |
+| [`ReducerEventContext` type](#type-reducereventcontext) | Implements [`IDbContext`](##interface-idbcontext) for [reducer callbacks](#observe-and-invoke-reducers). |
+| [`SubscriptionEventContext` type](#type-subscriptioneventcontext) | Implements [`IDbContext`](##interface-idbcontext) for [subscription callbacks](#subscribe-to-queries). |
+| [`ErrorContext` type](#type-errorcontext)                   | Implements [`IDbContext`](##interface-idbcontext) for error-related callbacks. |
+| [Access the client cache](#access-the-client-cache)         | Access to your local view of the database. |
+| [Observe and invoke reducers](#observe-and-invoke-reducers) | Send requests to the database to run reducers, and register callbacks to run when notified of reducers. |
+| [Identify a client](#identify-a-client)                     | Types for identifying users and client connections. |
 
-- [The SpacetimeDB C# client SDK](#the-spacetimedb-c-client-sdk)
-  - [Table of Contents](#table-of-contents)
-  - [Install the SDK](#install-the-sdk)
-    - [Using the `dotnet` CLI tool](#using-the-dotnet-cli-tool)
-    - [Using Unity](#using-unity)
-  - [Generate module bindings](#generate-module-bindings)
-  - [Initialization](#initialization)
-    - [Property `SpacetimeDBClient.instance`](#property-spacetimedbclientinstance)
-    - [Class `NetworkManager`](#class-networkmanager)
-    - [Method `SpacetimeDBClient.Connect`](#method-spacetimedbclientconnect)
-    - [Event `SpacetimeDBClient.onIdentityReceived`](#event-spacetimedbclientonidentityreceived)
-    - [Event `SpacetimeDBClient.onConnect`](#event-spacetimedbclientonconnect)
-  - [Subscribe to queries](#subscribe-to-queries)
-    - [Method `SpacetimeDBClient.Subscribe`](#method-spacetimedbclientsubscribe)
-    - [Event `SpacetimeDBClient.onSubscriptionApplied`](#event-spacetimedbclientonsubscriptionapplied)
-    - [Method \[`SpacetimeDBClient.OneOffQuery`\]](#method-spacetimedbclientoneoffquery)
-  - [View rows of subscribed tables](#view-rows-of-subscribed-tables)
-    - [Class `{TABLE}`](#class-table)
-      - [Static Method `{TABLE}.Iter`](#static-method-tableiter)
-      - [Static Method `{TABLE}.FilterBy{COLUMN}`](#static-method-tablefilterbycolumn)
-      - [Static Method `{TABLE}.FindBy{COLUMN}`](#static-method-tablefindbycolumn)
-      - [Static Method `{TABLE}.Count`](#static-method-tablecount)
-      - [Static Event `{TABLE}.OnInsert`](#static-event-tableoninsert)
-      - [Static Event `{TABLE}.OnBeforeDelete`](#static-event-tableonbeforedelete)
-      - [Static Event `{TABLE}.OnDelete`](#static-event-tableondelete)
-      - [Static Event `{TABLE}.OnUpdate`](#static-event-tableonupdate)
-  - [Observe and invoke reducers](#observe-and-invoke-reducers)
-    - [Class `Reducer`](#class-reducer)
-      - [Static Method `Reducer.{REDUCER}`](#static-method-reducerreducer)
-      - [Static Event `Reducer.On{REDUCER}`](#static-event-reduceronreducer)
-    - [Class `ReducerEvent`](#class-reducerevent)
-      - [Enum `Status`](#enum-status)
-        - [Variant `Status.Committed`](#variant-statuscommitted)
-        - [Variant `Status.Failed`](#variant-statusfailed)
-        - [Variant `Status.OutOfEnergy`](#variant-statusoutofenergy)
-  - [Identity management](#identity-management)
-    - [Class `AuthToken`](#class-authtoken)
-      - [Static Method `AuthToken.Init`](#static-method-authtokeninit)
-      - [Static Property `AuthToken.Token`](#static-property-authtokentoken)
-      - [Static Method `AuthToken.SaveToken`](#static-method-authtokensavetoken)
-    - [Class `Identity`](#class-identity)
-  - [Customizing logging](#customizing-logging)
-    - [Interface `ISpacetimeDBLogger`](#interface-ispacetimedblogger)
-    - [Class `ConsoleLogger`](#class-consolelogger)
-    - [Class `UnityDebugLogger`](#class-unitydebuglogger)
-
-## Install the SDK
+## Project setup
 
 ### Using the `dotnet` CLI tool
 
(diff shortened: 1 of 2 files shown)


docs/sdks/c-sharp/index.md
```diff
@@ -81,853 +47,813 @@ spacetime generate --lang cs --out-dir module_bindings --project-path PATH-TO-MO
...
+Call the `ctx.Db()` when operating in generic contexts.
 
-    // Our module contains a table named "Loot"
-    Loot.OnInsert += Loot_OnInsert;
+#### Example
 
-    SpacetimeDBClient.instance.Connect(/* ... */);
-}
+```csharp
+// TODO
```
joshua-spacetime, line 244: Rust sdk reference has an example of calling `ctx.Db()`

docs/sdks/c-sharp/index.md
```diff
@@ -81,853 +47,813 @@ spacetime generate --lang cs --out-dir module_bindings --project-path PATH-TO-MO
...
 
-```cs
-namespace SpacetimeDB {
+#### Example
 
-class SpacetimeDBClient {
-    public event Action onSubscriptionApplied;
-}
+```csharp
+// TODO
```
joshua-spacetime, line 261: Rust sdk has an example of calling `ctx.Reducers()`

docs/sdks/c-sharp/index.md
```diff
@@ -81,853 +47,813 @@ spacetime generate --lang cs --out-dir module_bindings --project-path PATH-TO-MO
...
+    public delegate void RowEventHandler(EventContext context, Row row);
+    public event RowEventHandler? OnInsert;
 }
 ```
 
-A unique public identifier for a user of a database.
+The `OnInsert` callback runs whenever a new row is inserted into the client cache, either when applying a subscription or being notified of a transaction. The passed [`EventContext`](#type-eventcontext) contains an [`Event`](#record-event) which can identify the change which caused the insertion, and also allows the callback to interact with the connection, inspect the client cache and invoke reducers. Newly registered or canceled callbacks do not take effect until the following event.
 
-<!-- FIXME: this is no longer accurate; `Identity` columns are properly `Identity`-type. -->
+TODO: Add example of registering and unregistering callback
```
joshua-spacetime, line 803: I.e. `.OnInsert +=` `.OnInsert -=`

docs/sdks/c-sharp/index.md
```diff
@@ -81,853 +47,813 @@ spacetime generate --lang cs --out-dir module_bindings --project-path PATH-TO-MO
...
 }
 ```
 
-An opaque identifier for a client connection to a database, intended to differentiate between connections from the same [`Identity`](#class-identity).
-
-## Customizing logging
+The `OnDelete` callback runs whenever a previously-resident row is deleted from the client cache. Newly registered or canceled callbacks do not take effect until the following event.
 
-The SpacetimeDB C# SDK performs internal logging.
+TODO: Add example of registering and unregistering callback
```
joshua-spacetime, line 817: I.e. `.OnDelete +=` `.OnDelete -=`

docs/sdks/c-sharp/index.md
```diff
@@ -81,853 +47,813 @@ spacetime generate --lang cs --out-dir module_bindings --project-path PATH-TO-MO
...
+}
+```
 
-### Interface `ISpacetimeDBLogger`
+The `OnUpdate` callback runs whenever an already-resident row in the client cache is updated, i.e. replaced with a new row that has the same primary key. Newly registered or canceled callbacks do not take effect until the following event.
 
-```cs
-namespace SpacetimeDB
-{
+TODO: Add example of registering and unregistering callback
```
joshua-spacetime, line 831: I.e. `.OnUpdate +=` `.OnUpdate -=`

docs/sdks/c-sharp/index.md
```diff
@@ -81,853 +47,813 @@ spacetime generate --lang cs --out-dir module_bindings --project-path PATH-TO-MO
...
+```csharp
+class RemoteTableHandle
+{
+    public delegate void RowEventHandler(EventContext context, Row row);
+    public event RowEventHandler? OnUpdate;
+}
+```
 
-### Interface `ISpacetimeDBLogger`
+The `OnUpdate` callback runs whenever an already-resident row in the client cache is updated, i.e. replaced with a new row that has the same primary key. Newly registered or canceled callbacks do not take effect until the following event.
```
joshua-spacetime, line 829: Can you set this if the table doesn't have a primary key?

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,859 @@
...
+    public DbView Db();
+}
+```
+
+Call the `ctx.Db()` when operating in generic contexts.
+
+#### Example
+
+```csharp
+// TODO
```
joshua-spacetime, line 244: Rust sdk reference has an example of calling `ctx.Db()`

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,859 @@
...
+    public RemoteReducers Reducers();
+}
+```
+
+Call the `ctx.Reducers()` when operating in generic contexts.
+
+#### Example
+
+```csharp
+// TODO
```
joshua-spacetime, line 261: Rust sdk has an example of calling `ctx.Reducers()`

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,859 @@
...
+class RemoteTableHandle
+{
+    public delegate void RowEventHandler(EventContext context, Row row);
+    public event RowEventHandler? OnInsert;
+}
+```
+
+The `OnInsert` callback runs whenever a new row is inserted into the client cache, either when applying a subscription or being notified of a transaction. The passed [`EventContext`](#type-eventcontext) contains an [`Event`](#record-event) which can identify the change which caused the insertion, and also allows the callback to interact with the connection, inspect the client cache and invoke reducers. Newly registered or canceled callbacks do not take effect until the following event.
+
+TODO: Add example of registering and unregistering callback
```
joshua-spacetime, line 803: I.e. `.OnInsert +=` `.OnInsert -=`

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,859 @@
...
+class RemoteTableHandle
+{
+    public delegate void RowEventHandler(EventContext context, Row row);
+    public event RowEventHandler? OnDelete;
+}
+```
+
+The `OnDelete` callback runs whenever a previously-resident row is deleted from the client cache. Newly registered or canceled callbacks do not take effect until the following event.
+
+TODO: Add example of registering and unregistering callback
```
joshua-spacetime, line 817: I.e. `.OnDelete +=` `.OnDelete -=`

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,859 @@
...
+class RemoteTableHandle
+{
+    public delegate void RowEventHandler(EventContext context, Row row);
+    public event RowEventHandler? OnUpdate;
+}
+```
+
+The `OnUpdate` callback runs whenever an already-resident row in the client cache is updated, i.e. replaced with a new row that has the same primary key. Newly registered or canceled callbacks do not take effect until the following event.
+
+TODO: Add example of registering and unregistering callback
```
joshua-spacetime, line 831: I.e. `.OnUpdate +=` `.OnUpdate -=`

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,859 @@
...
+
+```csharp
+class RemoteTableHandle
+{
+    public delegate void RowEventHandler(EventContext context, Row row);
+    public event RowEventHandler? OnUpdate;
+}
+```
+
+The `OnUpdate` callback runs whenever an already-resident row in the client cache is updated, i.e. replaced with a new row that has the same primary key. Newly registered or canceled callbacks do not take effect until the following event.
```
joshua-spacetime, line 829: Can you set this if the table doesn't have a primary key?

docs/sdks/c-sharp/index-v2.md
joshua-spacetime: Just point to quickstart

docs/sdks/c-sharp/index-v2.md
joshua-spacetime: Just point to quickstart

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,860 @@
...
+
+#### Method `FrameTick`
+
+```csharp
+class DbConnection {
+    public void FrameTick();
+}
+```
+
+`FrameTick` will advance the connection until no work remains or until it is disconnected, then return rather than blocking. Games might arrange for this message to be called every frame.
```
kazimuth, line 176:
```suggestion
`FrameTick` will advance the connection until no work remains or until it is disconnected, then return rather than blocking. Games might arrange for this message to be called every frame.

It is not advised to run `FrameTick` on a background thread, since it modifies [`dbConnection.Db`](#property-db). If main thread code is also accessing the `Db`, it may observe data races when `FrameTick` runs on another thread.

(Note that the SDK already does most of the work for parsing messages on a background thread. `FrameTick()` does the minimal amount of work needed to apply updates to the `Db`.)
```

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,860 @@
...
+##### Method `Subscribe`
+
+```csharp
+class SubscriptionBuilder
+{
+    public SubscriptionHandle Subscribe(string[] querySqls);
+}
+```
+
+Subscribe to a set of queries. `queries` should be a string or an array, vec or slice of strings.
```
kazimuth, line 342:
```suggestion
Subscribe to a set of queries. `queries` should be an array of SQL query strings.
```

docs/sdks/c-sharp/index-v2.md
kazimuth: Yes, but it does nothing.

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,860 @@
...
+}
+```
+
+The `OnUpdate` callback runs whenever an already-resident row in the client cache is updated, i.e. replaced with a new row that has the same primary key. Newly registered or canceled callbacks do not take effect until the following event.
+
+TODO: Add example of registering and unregistering callback
+
+### Unique constraint index access
+
+For each unique constraint on a table, its table handle has a property which is a unique index handle and whose name is the unique column name. This unique index handle has a method `.Find(Column: value)`. If a `Row` with `value` in the unique column is resident in the client cache, `.Find` returns it. Otherwise it returns null.
```
kazimuth, line 842: This needs an example, also should say `.Find(Column value)`.

Module-side I just unique indexes as a class. 

https://github.com/clockworklabs/spacetime-docs/pull/193/files#diff-84b02ce60e346c81f89c934fbaa3096d08eac9ee7c3f2cac7372efa3127ef128R450

Module-side we kind of have to fake it --the base class doesn't actually have all the methods I pretend it does, due to nullable generic struct/class problems. I don't know if that's the case SDK-side.

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,860 @@
...
+
+### BTree index access
+
+For each btree index defined on a remote table, its corresponding table handle has a property which is a btree index handle and whose name is the name of the index. This index handle has a method `IEnumerable<Row> Filter(Column value)` which will return `Row`s with `value` in the indexed `Column`, if there are any in the cache.
```
kazimuth, line 846: Similarly, an example here would be nice.

docs/sdks/c-sharp/index-v2.md
```diff
@@ -0,0 +1,860 @@
...
+
+- An invoke method, whose name is the reducer's name converted to snake case, like `set_name`. This requests that the module run the reducer.
+- A callback registation method, whose name is prefixed with `on_`, like `on_set_name`. This registers a callback to run whenever we are notified that the reducer ran, including successfully committed runs and runs we requested which failed. This method returns a callback id, which can be passed to the callback remove method.
+- A callback remove method, whose name is prefixed with `remove_on_`, like `remove_on_set_name`. This cancels a callback previously registered via the callback registration method.
+
+## Identify a client
+
+### Type `Identity`
+
+A unique public identifier for a client connected to a database.
```
kazimuth, line 862: May want methods, or alternatively just link to the module-side `Identity` docs once I write those up...

kazimuth (commented):
Looks great.

There's some stuff shared between the module library and client SDK, specifically `Identity`, `ConnectionId`, and `TaggedEnum`. We may want to just link to the module documentation here, or copy it over. I haven't fully documented `Identity` or `ConnectionId` yet module-side.

docs/sdks/c-sharp/index-v2.md
kazimuth: Yes, but it does nothing.

docs/sdks/c-sharp/index-v2.md
joshua-spacetime: I added a couple comments for linking to the module docs once those are ready.

kazimuth: We'll also want `Timestamp` and `TimeDuration`. I don't think `ScheduleAt` is necessary here.

kazimuth (approved):
I am happy with this, although we'll need to merge the other ref first.

docs/sdks/c-sharp/index-v2.md
gefjon: I'm making a follow-up ticket. `Timestamp` and `TimeDuration` aren't documented in any of the SDKs currently. We'll get to them in a fast-follow.

Issue #146, opened 2025-01-27
Deploying: add "self-hosted" page


bfops:
This is where we'd put things like how to set up a reverse proxy with route filtering. Tell people to block the `/publish` route from external clients, and to publish only by ssh'ing into the machine and doing a `spacetime publish -s local`.

Tyler asserts this is a P1.

bfops:
@cloutiertyler thinks @kurtismullins is equipped to do this. (@drogus originally set this up, but he has a lot of tickets)

gefjon:
Could be published a few days after the announcement, but not much later than that.

bfops:
Closing because this issue is marked Done

PR #206, opened 2025-03-04
Self-Hosted guide

jdetter:
Solves https://github.com/clockworklabs/spacetime-docs/issues/146

Rendered: https://github.com/clockworklabs/spacetime-docs/blob/jdetter/spacetimedb-standalone-guide/docs/deploying/spacetimedb-standalone.md

There is a small issue where if you try to use `--root-dir` during install then there's issues executing the binaries. It seems like there's a bug in the `spacetime` CLI that prevents this from working so instead I just have the user install to the default installation directories.

## Testing

I've gone through the entire tutorial myself including issuing an SSL certificate. I deployed a blackholio here:

```
url: https://tutorial-test.spacetimedb.com
name: blackholio
```

TODO: tear down the machine and the DNS entry used for testing.


docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
+# Self Hosting SpacetimeDB
+
+This tutorial will guide you through setting up SpacetimeDB on an Ubuntu 24.04 server, securing it with HTTPS using Nginx and Let's Encrypt, and configuring a systemd service to keep it running.
+
+## Prerequisites
+- A fresh Ubuntu 24.04 server (VM or cloud instance of your choice)
+- A domain name (e.g., `example.com`)
+- `sudo` privileges on the server
+
+Also, join the community! We have an official discord server: https://discord.gg/spacetimedb
```
gefjon, line 10: This feels weird and breaks up the flow of the tutorial. We tell people to join the Discord plenty; we don't need to do it here too.

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Step 1: Create a Dedicated User for SpacetimeDB
+For security purposes, create a dedicated `spacetimedb` user to run SpacetimeDB with a home directory at `/stdb`:
+
+```sh
+sudo mkdir /stdb
+sudo useradd --system spacetimedb
+sudo chown -R spacetimedb:spacetimedb /stdb
+```
```
gefjon, line 19: Does this actually set the user's homedir? I think you have to set `-d` or `--home` to `useradd` to do that. It's not clear to me that you actually want to set the user's homedir to be `/stdb`; I think you just want to create `/stdb` as the SpacetimeDB data dir. If that's the case, you shouldn't use the phrase "home directory" to describe this.

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+
+Switch to the new user and install SpacetimeDB:
+
+```sh
+sudo -u spacetimedb bash -c 'curl -sSf https://install.spacetimedb.com | sh -s -- --root-dir /stdb --yes'
+```
```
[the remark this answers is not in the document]
gefjon: Eh, rustup.rs tells people to pipe `curl` into `sh`, and they don't have any justification or alternative. I think this is fine.

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+    # SpacetimeDB server in production.
+    # location /v1/publish {
+        # allow 127.0.0.1;
+        # deny all;
+        # proxy_pass http://localhost:3000;
+        # proxy_http_version 1.1;
+        # proxy_set_header Upgrade $http_upgrade;
+        # proxy_set_header Connection "Upgrade";
+        # proxy_set_header Host $host;
+    # }
```
gefjon, line 107: I would have this uncommented, and have the comment suggest that the reader might want to comment this out if they have controlled access to the server some other way. This tutorial is about deploying SpacetimeDB-standalone for production, so it should show the production setup by default.

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+spacetime publish -s self-hosted <module-name>
+```
+
+If you have uncommented the `/v1/publish` restriction in Step 3 then you won't be able to publish to this instance unless you copy your module to the host first and then publish. We recommend something like this:
+
+```bash
+spacetime build
+scp target/wasm32-unknown-unknown/release/spacetime_module.wasm ubuntu@<host>:/home/ubuntu/
+ssh ubuntu@<host> spacetime publish -s local --bin-path spacetime_module.wasm <module-name>
+```
```
gefjon, line 182: Again, I would position the restricted mode as the default.

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+```bash
+spacetime publish -s self-hosted <module-name>
+```
+
+If you have uncommented the `/v1/publish` restriction in Step 3 then you won't be able to publish to this instance unless you copy your module to the host first and then publish. We recommend something like this:
+
+```bash
+spacetime build
+scp target/wasm32-unknown-unknown/release/spacetime_module.wasm ubuntu@<host>:/home/ubuntu/
+ssh ubuntu@<host> spacetime publish -s local --bin-path spacetime_module.wasm <module-name>
```
gefjon, line 181: If you're gonna carve out `<host>`, do you also want to say `<user>` instead of `ubuntu`?

nav.ts
```diff
@@ -91,6 +91,7 @@ const nav: Nav = {
 
     section('How To'),
     page('Incremental Migrations', 'how-to/incremental-migrations', 'how-to/incremental-migrations.md'),
+    page('Self-Hosting SpacetimeDB', 'how-to/spacetimedb-standalone', 'how-to/spacetimedb-standalone.md'),
```
gefjon, line 94: This should go in the "Deploying" section, around line 36.

gefjon (changes requested):

docs/how-to/spacetimedb-standalone.md
jdetter: I think the CTA is important here - I think more people will use this guide as it's been frequently requested. This install is also slightly more complicated and I want people to join the discord if they run into issues. We could rephrase to something like, "If you run into issues, reach out to us in Discord! <link>" What do you think?

docs/how-to/spacetimedb-standalone.md
jdetter: Oh sorry for the confusion. I actually rewrote this section to not set the home directory, I will fix the text above.

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Prerequisites
+- A fresh Ubuntu 24.04 server (VM or cloud instance of your choice)
+- A domain name (e.g., `example.com`)
+- `sudo` privileges on the server
+
+Also, join the community! We have an official discord server: https://discord.gg/spacetimedb
+
+## Step 1: Create a Dedicated User for SpacetimeDB
+For security purposes, create a dedicated `spacetimedb` user to run SpacetimeDB with a home directory at `/stdb`:
```
jdetter, line 13:
```suggestion
For security purposes, create a dedicated `spacetimedb` user to run SpacetimeDB:
```

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+WorkingDirectory=/stdb
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Enable and start the service:
+
+```sh
+sudo systemctl daemon-reload
```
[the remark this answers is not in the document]
jdetter: Good point, probably not

jdetter, line 54:
```suggestion
```

docs/how-to/spacetimedb-standalone.md
jdetter: I think we should keep it as `ubuntu` mainly because I don't want people getting confused and uploading the module under the `spacetime` user (which that user does not have a home directory). What do you think?

docs/how-to/spacetimedb-standalone.md
jdetter: I originally had it uncommented and then I chickened out 😆 I agree with your analysis, I'll change it 👍

docs/how-to/spacetimedb-standalone.md
jdetter: I agree, updated 👍

docs/how-to/spacetimedb-standalone.md
```diff
@@ -0,0 +1,250 @@
...
+        # proxy_pass http://localhost:3000;
+        # proxy_http_version 1.1;
+        # proxy_set_header Upgrade $http_upgrade;
+        # proxy_set_header Connection "Upgrade";
+        # proxy_set_header Host $host;
+    # }
+}
+```
+
+This configuration contains a restriction to the `/v1/publish` route. This restriction makes it so that you can only publish to the database if you're publishing from a local connection on the host. You should uncomment that section and then run `sudo systemctl nginx restart` in order to apply that config change.
```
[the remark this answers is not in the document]
jdetter: this bit got deleted but thanks for the edit 👍

docs/deploying/spacetimedb-standalone.md
```diff
@@ -0,0 +1,242 @@
+# Self Hosting SpacetimeDB
+
+This tutorial will guide you through setting up SpacetimeDB on an Ubuntu 24.04 server, securing it with HTTPS using Nginx and Let's Encrypt, and configuring a systemd service to keep it running.
+
+## Prerequisites
+- A fresh Ubuntu 24.04 server (VM or cloud instance of your choice)
+- A domain name (e.g., `example.com`)
+- `sudo` privileges on the server
+
+Also, join the community! We have an official discord server: https://discord.gg/spacetimedb
```
jdetter, line 10: For consistency we shall remove this
```suggestion
```

docs/deploying/spacetimedb-standalone.md
```diff
@@ -0,0 +1,241 @@
+# Self Hosting SpacetimeDB
+
+This tutorial will guide you through setting up SpacetimeDB on an Ubuntu 24.04 server, securing it with HTTPS using Nginx and Let's Encrypt, and configuring a systemd service to keep it running.
+
+## Prerequisites
+- A fresh Ubuntu 24.04 server (VM or cloud instance of your choice)
+- A domain name (e.g., `example.com`)
+- `sudo` privileges on the server
+
```
jdetter, line 9:
```suggestion
```

gefjon (approved):

docs/deploying/spacetimedb-standalone.md
```diff
@@ -0,0 +1,240 @@
...
+## Step 1: Create a Dedicated User for SpacetimeDB
+For security purposes, create a dedicated `spacetimedb` user to run SpacetimeDB:
+
+```sh
+sudo mkdir /stdb
+sudo useradd --system spacetimedb
+sudo chown -R spacetimedb:spacetimedb /stdb
+```
+
+Switch to the new user and install SpacetimeDB:
```
jdetter, line 19:
```suggestion
Install SpacetimeDB as the new user:
```

PR #216, opened 2025-03-05
Adjust the BSATN doc to fit reality better

Centril:
This is how it's actually defined in the sats module these days.


rekhoff (approved):
This update is much-more clear and current. Approved!

PR #221, opened 2025-03-07
Updated to correct old timestamp format.

rekhoff:
Error in documentation, for Blackholio quickstart, that used the older timestamp code. Corrected to bring it up to date with the repo example found here: https://github.com/clockworklabs/Blackholio/blob/8d668332884b366587d4e63606dfa4f4a9899f33/server-rust/src/lib.rs#L138


cloutiertyler (approved):
Thank you! This is syncing with what already exists in the Blackholio repo

PR #253, opened 2025-03-09
Table name is 'messages' not 'Messages'

kaneviggers:
Was just following the QuickStart and came across this.

Changing the command in the quickstart guide from 
`spacetime sql quickstart-chat "SELECT * FROM Message"`
to
`spacetime sql quickstart-chat "SELECT * FROM message"`

As the `message` table in the code block starts with a lowercase `m`


bfops:
Thank you for fixing this!

bfops (approved):

PR #255, opened 2025-03-10
fix(ts-sdk): correct the function signature of onConnectError

cjodo:


bfops (approved):
Looks right to me, thank you!

PR #264, opened 2025-03-11
Small fix on part-3.md

AdielMag:
The instruction on c# stated that it would show an error because `SpawnFoodTimer` was not in the method signature but in the initial Implementation it is, so I removed it.


bfops (approved):
Good catch, thank you!

PR #270, opened 2025-03-14
CI - Add release branch check

bfops:
This check will fail if a PR is opened onto the `release` branch from a branch that doesn't look like `release-*`.

This is similar to the check we have in the C#/Unity SDK (https://github.com/clockworklabs/com.clockworklabs.spacetimedbsdk/blob/release/latest/.github/workflows/check-pr-base.yml).


jdetter (approved):
Looks good, thanks Zeke 👍

PR #278, opened 2025-03-28
Add "Subscription Semantics" page

gefjon:
We've gotten several questions in the public Discord about the semantics of subscriptions in the SDKs. I did a brain-dump about them, Tyler fed it into ChatGPT, I touched up the result a bit, and here it is. Some day we probably want to rewrite this to read less like AI slop. But for now it's probably fine.


jdetter (approved):
Well written, thank you @gefjon !

PR #279, opened 2025-04-01
Fix typo in DBVector2 struct name in part-3.md

rlneumiller:
Fix for typo in code section (DbVector -> DBVector2) that would result in compile error


bfops:
Thanks for opening a PR! Are you sure that this is correct though? We seem to use `DbVector2` everywhere else in the docs, and I don't see any other mentions of `DBVector2` :thinking:

rlneumiller:
Oops - looks like I had it the wrong-way-round.  Excuses available upon request.

PR #280, opened 2025-04-02
TS quickstart: Install latest SDK, not specific version

gefjon:
The version listed here was outdated and included bugs. I don't even know why we'd recommend a specific version, instead of just telling people to install the latest release.

[From a report on Discord](https://discord.com/channels/1037340874172014652/1356648963025141901/1356933593422630922).


jdetter (approved):
I believe this will install the version that we have configured on the npm website as the "latest release". We've added an instruction to the release checklist to update this version after doing a release so this should work as expected👍

PR #291, opened 2025-04-11
Reference docs for row level security

joshua-spacetime:


docs/rls/index.md
```diff
@@ -0,0 +1,258 @@
...
+spacetimedb = { version = "...", features = ["unstable"] }
+```
+
+### C#
+
+To enable RLS in C#, include the following preprocessor directive at the top of your module files:
+
+```csharp
+#pragma warning disable STDB_UNSTABLE
+```
```
gefjon, line 24: Should we use the language-switcher widget here too?

docs/rls/index.md
```diff
@@ -0,0 +1,258 @@
...
+:::server-rust
+RLS rules are expressed in SQL and declared as constants of type `Filter`.
+
+```rust
+use spacetimedb::{client_visibility_filter, Filter};
+
+/// A client can only see their user
+#[client_visibility_filter]
+const USERS_FILTER: Filter = Filter::Sql(
+    "SELECT * FROM users WHERE identity = :sender"
```
gefjon, line 37: I believe we use singular table names by convention, so this should be `user`, not `users`.

docs/rls/index.md
```diff
@@ -0,0 +1,258 @@
...
+#pragma warning disable STDB_UNSTABLE
+
+public partial class Module
+{
+    /// <summary>
+    /// A client can only see their user.
+    /// </summary>
+    [SpacetimeDB.ClientVisibilityFilter]
+    public static readonly Filter USERS_FILTER = new Filter.Sql(
+        "SELECT * FROM users WHERE identity = :sender"
```
gefjon, line 56: Ditto, singular table name.

docs/rls/index.md
```diff
@@ -0,0 +1,258 @@
...
+    [SpacetimeDB.ClientVisibilityFilter]
+    public static readonly Filter PLAYERS_FILTER = new Filter.Sql(
+        "SELECT p.* FROM users u JOIN players p ON u.id = p.id"
+    );
+}
+```
+:::
+
+And while self-joins are allowed, in general RLS rules cannot be self-referential,
+as this would result in infinite recursion.
```
gefjon, line 199: I'm not sure I understand this distinction. Do you have an example of a query that's not valid for this reason?

gefjon (commented):
All the table names here should be singular rather than plural. Structure and content are great!

docs/rls/index.md
joshua-spacetime: Typically you don't see a table named `user` since `USER` is a reserved keyword in SQL. And I couldn't think of a better name for the table in the context of my examples. I'm open to suggestions though.

Also FWIW, using singular names is not a universal convention. Postgres tends to use plural names in their documentation.

docs/rls/index.md
joshua-spacetime: Yes I can add one.

docs/rls/index.md
gefjon: > Also FWIW, using singular names is not a universal convention. Postgres tends to use plural names in their documentation.

All I care about is that we maintain a convention. @cloutiertyler at some point declared that the convention would be singular names, and since then I've been upholding it.

I would recommend using `account` as the name for this table.

cloutiertyler: We have a user table in postgres for the website. We just have to quote the name `"user"`.

gefjon (approved):

PR #298, opened 2025-04-14
Fix typo in unity/part-2.md

waaverecords:
`is only every called` to `is only ever called`


bfops (approved):
LGTM! Thanks for fixing this.

PR #301, opened 2025-04-15
Explain primary_key and the distinction between primary_key and unique #271

### LINKED ISSUE #271: Explain `primary_key` and the distinction between `primary_key` and `unique`
> In SpacetimeDB, the only difference between unique and primary_key is that the client SDKs assign special semantics to primary_key columns. Specifically, the on_update callback fires for a row when a single transaction deletes an old row with a certain primary key, and then inserts a new row with that same primary key. Tables without a primary key column will never observe an on_update event. (Also, a table can only have at most one primary key column, but may have as many unique columns as it wants.) Otherwise, the two constraints are equivalent.

Include the above in the docs

tamaro-skaljic:
Fixes #271


bfops:
Thank you for this! I think we want to move this into a different area of the docs, since it's not really specific to our SQL interface, but we're not entirely sure where it should go. Maybe the quickstarts or reference docs..

PR #302, opened 2025-04-15
Document sql reserved identifiers #266

### LINKED ISSUE #266: Document sql reserved identifiers
While we document the Spacetime SQL grammar, we don't mention that we use a postgres compatible parser, which means reserved identifiers in postgres are automatically reserved in Spacetime SQL. Users can always get around this by using quoted identifiers, but perhaps this should be better documented. A user brought this up in https://github.com/clockworklabs/SpacetimeDB/issues/2438.

tamaro-skaljic:
Fixes #266


joshua-spacetime (approved):
Thank you @tamaro-skaljic!

PR #310, opened 2025-04-15
fix wrong quote usage and copy the "start the server" section from c# quickstart to rust quickstart #41

### LINKED ISSUE #41: Two important things should be changed in the Rust docs
First, the command to send a message located [here](https://spacetimedb.com/docs/modules/rust/quickstart#call-reducers) does not work correctly (at least on Windows). With single quotes, the text sent to the database is `'Hello,`. In order to send the whole `Hello, World!` message, you have to use double quotes. The whole command should be `spacetime call <module-name> send_message "Hello, World!"`.

Second, using `spacetime start` is never mentioned as a prerequisite. If you follow the guide without doing that, you get this error message:
```
>spacetime publish --project-path server quickstart-chat-module
info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date
checking crate with spacetimedb's clippy configuration
    Checking quickstart-chat-module v0.1.0
    Finished release [optimized] target(s) in 0.19s
    Finished release [optimized] target(s) in 0.08s
Optimising module with wasm-opt...
Uploading to local => http://127.0.0.1:3000
Error: error sending request for url (http://127.0.0.1:3000/identity): error trying to connect: tcp connect error: No connection could be made because the target machine actively refused it. (os error 10061)

Caused by:
    0: error trying to connect: tcp connect error: No connection could be made because the target machine actively refused it. (os error 10061)
    1: tcp connect error: No connection could be made because the target machine actively refused it. (os error 10061)
    2: No connection could be made because the target machine actively refused it. (os error 10061)
```
which is unclear as well. It would be good to have a check before the spacetime CLI tries to connect to the database to determine if it is up in the first place.

tamaro-skaljic:
Fixes #41


docs/modules/rust/quickstart.md
```diff
@@ -239,7 +243,7 @@ spacetime publish --project-path server quickstart-chat
 You can use the CLI (command line interface) to run reducers. The arguments to the reducer are passed in JSON format.
 
 ```bash
-spacetime call quickstart-chat send_message 'Hello, World!'
+spacetime call quickstart-chat send_message "Hello, World!"
```
gefjon, line 246: Why is this change necessary?

bfops: I don't particularly love this, but since this still works in `bash`, and since this is a quickstart, I think it's reasonable to make this quickstart support other shells (even if we don't commit to supporting them everywhere).

bfops (approved):
I suspect other users will hit the same issues and it still works with `bash`

PR #311, opened 2025-04-15
Deploying: Maincloud: add section for connecting via SDKs

bfops:
Update the `Deploying: Maincloud` page to include snippets for building a corresponding `DbConnection` in our client languages.


gefjon (commented):
I think this is overkill in a way that distracts from the point. The only thing that's different is the `with_uri`/`withUri`/`WithUri` call, so I  think we should aim to show as close to only that as possible.

I'd also like to get a clear sentence like, "The URI to connect to Maincloud is `https://maincloud.spacetimedb.com`.

bfops:
How's this?

gefjon (approved):

PR #313, opened 2025-04-16
Fixed RLS Sql examples

Lethalchip:
LN99, LN124, LN150 & LN181 were missing the `ON account.identity = admin.identity` part.

LN211 & LN230 accidentally referenced `u.id` instead of `a.id`- I'm assuming because it was `user` previously, instead of `account`, and was missed.

### DIFF
--- a/docs/rls/index.md
+++ b/docs/rls/index.md
@@ -96,7 +96,7 @@ const ACCOUNT_FILTER: Filter = Filter::Sql(
 /// An admin can see all accounts
 #[client_visibility_filter]
 const ACCOUNT_FILTER_FOR_ADMINS: Filter = Filter::Sql(
-    "SELECT account.* FROM account JOIN admin WHERE admin.identity = :sender"
+    "SELECT account.* FROM account JOIN admin ON account.identity = admin.identity WHERE admin.identity = :sender"
 );
 ```
 :::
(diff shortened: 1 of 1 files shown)


docs/rls/index.md
```diff
@@ -208,7 +208,7 @@ use spacetimedb::{client_visibility_filter, Filter};
 const PLAYER_FILTER: Filter = Filter::Sql("
     SELECT q.*
     FROM account a
-    JOIN player p ON u.id = p.id
+    JOIN player p ON a.id = p.id
```
joshua-spacetime, line 211: Thanks for catching this!

docs/rls/index.md
```diff
@@ -96,7 +96,7 @@ const ACCOUNT_FILTER: Filter = Filter::Sql(
 /// An admin can see all accounts
 #[client_visibility_filter]
 const ACCOUNT_FILTER_FOR_ADMINS: Filter = Filter::Sql(
-    "SELECT account.* FROM account JOIN admin WHERE admin.identity = :sender"
+    "SELECT account.* FROM account JOIN admin ON account.identity = admin.identity WHERE admin.identity = :sender"
```
joshua-spacetime, line 99: Perhaps the docs don't explain this sufficiently, but the reason there's no join condition is to allow an admin access to all accounts. Joining on `identity` means an admin can only see their own account information.

PR #277, opened 2025-03-27
Rename module to database where appropriate

tamaro-skaljic:
I recommend reviewing commit by commit.

I'm not sure about https://github.com/clockworklabs/spacetime-docs/pull/277/commits/20b2e7ba804861219a3a50a7dd2722d21d0d2130 and https://github.com/clockworklabs/spacetime-docs/pull/277/commits/6aa5d9c881c1e9cd8561eea85297a5dec61a8d81, you can revert them if it's wrong.

PR #317, opened 2025-04-17
Revert CLI reference to match what the CLI outputs

bfops:
https://github.com/clockworklabs/spacetime-docs/pull/277 updated the CLI reference section, which is intended to be autogenerated from the CLI itself.

In order to fix failing CI in SpacetimeDB, I am reverting the changes. If we want to make these changes again, we should make them in the CLI itself.


jdetter (approved):
These changes are more or less the same so I am fine with rolling these back. If it takes more than like 5-10 minutes I would *not* recommend that we update the CLI source files to apply this change to the auto-generated docs.

Thanks Zeke 👍

PR #318, opened 2025-04-17
CI - Check the CLI docs

bfops:


.github/workflows/check-cli-reference.yml
```diff
@@ -0,0 +1,45 @@
...
+          cd spacetime-docs
+          # This is needed because our website doesn't render markdown quite properly.
+          # See the README in spacetime-docs for more details.
+          sed -i'' -E 's!^(##) `(.*)`$!\1 \2!' docs/cli-reference.md
+          sed -i'' -E 's!^(######) \*\*(.*)\*\*$!\1 <b>\2</b>!' docs/cli-reference.md
+          git status
+          if git diff --exit-code HEAD; then
+            echo "No docs changes detected"
+          else
+            echo "It looks like the CLI docs have changed:"
```
jdetter, line 43: IMO We should add some extra details here, like if the docs *have* changed then how can they be updated?

jdetter:
Is your intention for this to become a required check?

.github/workflows/check-cli-reference.yml
bfops: How's this?

bfops:
> Is your intention for this to become a required check?

That's what I would do, yes.

jdetter (approved):
Updated comment looks good, thanks Zeke 👍

PR #323, opened 2025-04-20
fix links from rls docs to sql docs on website

tamaro-skaljic:
See https://spacetimedb.com/docs/rls


bfops (approved):
Good catch, thank you!

PR #324, opened 2025-04-21
Fix some broken links

bfops:
I grepped for `.md` inside our docs, and fixed all the links to `.md`. (Actually, there's one remaining, but it's to an external github repo and the link is correct).


bfops:
this was fixed in https://github.com/clockworklabs/spacetime-docs/pull/340

PR #327, opened 2025-04-21
Now noting that StdbModule.csproj is important

cloutiertyler:


jdetter (approved):
This is a good update, we have had a few users try to rename this file and it obviously caused problems when they went to publish.

PR #332, opened 2025-04-24
CI - Expand CI-triggering events

bfops:
Run (most of) our CI checks on commits to `master`. I also made them run on `merge_group`, in case we decide to use a merge queue in the future. I also restricted the release branch checks to only run on the release branch.

Once this is happening, we can re-enable these status checks as required on the `release` branch, which will give us a bit security against accidentally pushing a wrong/broken thing.


jdetter (approved):
Sounds good 👍

PR #333, opened 2025-04-25
Fix example that was still rust-ey in C# ref

gefjon:
This must have gotten missed when porting the Rust ref over to C#.


docs/modules/c-sharp/index.md
```diff
@@ -480,12 +480,12 @@ public static partial class Module {
     [Reducer]
     void Demo(ReducerContext ctx) {
         var idIndex = ctx.Db.user.Id;
-        var exampleUser = idIndex.find(357).unwrap();
-        exampleUser.dog_count += 5;
-        idIndex.update(exampleUser);
+        var exampleUser = idIndex.Find(357);
```
kazimuth, line 483: `User` is a struct, so `User?` is `Optional<User>`, and you have to use `.Value`. If it was a class, this code would be correct.

```suggestion
        var exampleUser = idIndex.Find(357).Value;
```

kazimuth (approved):
LGTM with fix

PR #334, opened 2025-04-26
Stronger wording to direct users to working unity versions

Lethalchip:
While Unity version `2021.2` onward may _technically_ be supported due to c# 9, it's likely to not going to work. I've tested half a dozen versions PRIOR to `2022.3.32f1` and none of them work out of the box. You will get errors like this: 
image

So, let's strongly word the version recommendation so that users are more inclined to use a working version of unity and not encounter these errors. 

ref: https://discord.com/channels/1037340874172014652/1037343198617538580/1365303831864410242
ref: https://discord.com/channels/1037340874172014652/1360001481893347560/1360001481893347560

### DIFF
--- a/docs/unity/index.md
+++ b/docs/unity/index.md
@@ -12,12 +12,12 @@ Our game, called [Blackhol.io](https://github.com/ClockworkLabs/Blackholio), wil
 
 This tutorial assumes that you have a basic understanding of the Unity Editor, using a command line terminal and programming. We'll give you some CLI commands to execute. If you are using Windows, we recommend using Git Bash or PowerShell. For Mac, we recommend Terminal.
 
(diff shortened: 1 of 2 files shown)


gefjon (commented):
Why are we hedging at all? "Recommend" is a needlessly weak word, when we could just say, "SpacetimeDB supports Unity version `2022.3.32f1` or later." We also don't need to tell people that earlier versions might work. We expect any version before the one we support to be at least partially broken, and so we should not give users any expectations of functionality. If we feel the need to make any comment at all about earlier versions (which I would not), it should be something like, "SpacetimeDB may be partially functional with earlier versions of SpacetimeDB, but key features will be broken or absent. The SpacetimeDB team is unable to offer support for issues related to earlier versions of Unity."

gefjon (approved):
Thanks a ton!

PR #335, opened 2025-04-27
Correcting Reducer Callback Information in llms.md

Lethalchip:
There is currently incorrect information within the reducer callback function in `llms.txt`: 

```typescript
private handleSendMessageResult(ctx: ReducerEventContext, messageText: string) {
    const wasOurCall = ctx.reducerEvent.callerIdentity.isEqual(this.identity);
    if (!wasOurCall) return; // Only care about our own calls here

    if (ctx.reducerEvent.status === Status.Committed) {
        console.log(`Our message "${messageText}" sent successfully.`);
    } else if (ctx.reducerEvent.status.isFailed()) {
        console.error(`Failed to send "${messageText}": ${ctx.reducerEvent.status.getFailedMessage()}`);
    }
}
```

Specifically: 
- `ctx.reducerEvent.callerIdentity.isEqual(this.identity)`
- `ctx.reducerEvent.status === Status.Committed`
- `ctx.reducerEvent.status.isFailed()`
- `ctx.reducerEvent.status.getFailedMessage()`

I've corrected the issues and ensured that this particular example compiles and works by adapting it to an existing project.

### DIFF
--- a/llms.md
+++ b/llms.md
@@ -2171,8 +2171,6 @@ private handleSendMessageResult(ctx: ReducerEventContext, messageText: string) {
     switch(ctx.event.status.tag) {
     case "Committed":
         console.log(`Our message "${messageText}" sent successfully.`);
-        // ctx.event.status.value is type DatabaseUpdate
-        // which has .tables which is an array of TableUpdates
         break;
     case "Failed":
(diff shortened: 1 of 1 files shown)


gefjon (approved):
Thanks so much!

gefjon:
Fixes #336

Issue #287, opened 2025-04-07
Fix instructions for blocking publishing from non-local IPs


bfops:
Our docs 1) use an outdated route for publishing, and 2) should be changed to have an allow list instead of a blocklist.

PR #338, opened 2025-04-30
Fix blocking publish route

jdetter:
This fixes an issue where the self-hosting docs were blocking an old API route. I've updated the docs to block all routes by default and then we can manually allow certain routes to go through (like subscribe, call, etc.).

In order to improve this flow I have created this ticket which will remove the requirement for a lot of this route blocking silliness: https://github.com/clockworklabs/spacetimedb/issues/2684

99.9% of users just want to block the publish route and it is more ergonomic to just allow users to specify their identity so that only they can create new databases.

## Testing

- [x] Default config allows anyone to publish
- [x] "Restricted" config denies external publishes:

```
Uploading to self-hosted => https://standalone-test.example.com
Publishing module...
Error: <html>
<head><title>403 Forbidden</title></head>
<body>
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx/1.24.0 (Ubuntu)</center>
</body>
</html>


Caused by:
    HTTP status client error (403 Forbidden) for url (https://standalone-test.example.com/v1/database)
```
- [x] Publishing locally on the self-hosted machine allows you to publish

## Follow-up

After this issue is fixed we need to update this documentation: https://github.com/clockworklabs/spacetimedb/issues/2684


docs/deploying/spacetimedb-standalone.md
```diff
@@ -82,25 +82,65 @@ server {
...
+    # Uncomment this section to optionally allow /v1/identity. This would allow remote users to create
+    # an identity for the purpose of connecting to your SpacetimeDB instance.
+    # NOTE: this is *not* required for the subscribe route to function.
+    # location /v1/identity {
+    #     proxy_pass http://localhost:3000;
+    #     proxy_http_version 1.1;
+    #     proxy_set_header Upgrade $http_upgrade;
+    #     proxy_set_header Connection "Upgrade";
+    #     proxy_set_header Host $host;
+    # }
```
jdetter, line 137: There are other routes that can be unblocked here as well but I think calling reducers, creating identities, and doing sql queries is going to cover the vast majority of use cases.

docs/deploying/spacetimedb-standalone.md
```diff
@@ -82,25 +82,67 @@ server {
...
+
+    # Block all other routes explicitly. Only localhost can use these routes.
+    # We would recommend uncommenting this section when you don't want remote users to
+    # publish to your database!
+    # location / {
+    #     allow 127.0.0.1;
+    #     deny all;
+    # }
```
bfops, line 145: we have lines below that say:
> This configuration contains a restriction to the `/v1/publish` route. This restriction makes it so that you can only publish to the database if you're publishing from a local connection on the host. 

Do we want to update that since that block isn't enabled by default anymore? (Maybe just removing it entirely, since I think the inline comments do a pretty good job).

jdetter:
Fixes https://github.com/clockworklabs/spacetime-docs/issues/287

docs/deploying/spacetimedb-standalone.md
jdetter: *Interesting*, I thought the default before was open. I will gladly changed the default to not be open and I'll update that paragraph 👍

docs/deploying/spacetimedb-standalone.md
```diff
@@ -82,24 +82,64 @@ server {
...
+    # users from creating new databases you should keep this section commented out. Otherwise, if you
+    # want to open it up (probably for dev environments) then you can uncomment this section.
+    #########################################
+    # location / {
+    #     proxy_pass http://localhost:3000;
+    #     proxy_http_version 1.1;
+    #     proxy_set_header Upgrade $http_upgrade;
+    #     proxy_set_header Connection "Upgrade";
+    #     proxy_set_header Host $host;
+    #     break;
```
bfops, line 96: nginx question - what is this `break` for?

docs/deploying/spacetimedb-standalone.md
bfops: (I think we still want to upgrade the paragraph at the very least to not say `/v1/publish` since we don't explicitly touch a `publish` route anymore)

bfops (changes requested):
See @gefjon's comment above

jdetter:
Tested in the following cases:

## Open Config

With the settings completely "open" where you can use any route I was able to publish + play blackholio with no problem

## Closed Config

With the closed config this is what you get when you publish remotely:
```
Publishing module...
Error: <html>
<head><title>403 Forbidden</title></head>
<body>
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx/1.24.0 (Ubuntu)</center>
</body>
</html>


Caused by:
    HTTP status client error (403 Forbidden) for url (https://standalone-test.<redacted>.com/v1/database/blackholio?clear=true)
```

The game still plays as expected.

## Typescript

I was not able to test this using typescript because the typescript quickstart (both the guide itself and the quickstart example in the typescript repository) is broken

jdetter:
@gefjon I unblocked `/v1/identity` by default now but I wasn't able to test because the example in the typescript quickstart sdk repo is broken and also the guide on the website is broken as well. I *think* it works though but I wasn't able to test it myself.

docs/deploying/spacetimedb-standalone.md
```diff
@@ -82,25 +82,64 @@ server {
     listen 80;
     server_name example.com;
 
-    location / {
+    #########################################
+    # By default SpacetimeDB is completely open so that anyone can publish to it. If you want to block
+    # users from creating new databases you should keep this section commented out. Otherwise, if you
+    # want to open it up (probably for dev environments) then you can uncomment this section.
```
bfops, line 88: I believe they would need to uncomment this section _and_ comment out the `location /` block below

bfops (approved):
LGTM, thank you!

jdetter:
Clarification on this, I followed the instructions in the README.md here and the typescript quickstart chat does in fact work: https://github.com/clockworklabs/spacetimedb-typescript-sdk/tree/main/examples/quickstart-chat

PR #340, opened 2025-05-01
Fix link checking

bfops:
The link checking had a few problems.

First of all, it wasn't checking "non-inline" links, i.e. where the link value itself is a "footnote" such as the ones fixed in https://github.com/clockworklabs/spacetime-docs/pull/324/files. I fixed this by using a genuine markdown parser.

Fixing revealed 9 new broken links, as well as some false positives due to the following issues:
- Links to the `/docs` root were erroneously being considered broken, because the slug is `/docs/index`
- Links to the homepage (`/`) were erroneously being considered broken, because `/` wasn't in the list of pages to skip.
- Links with trailing `/` were erroneously being considered broken, because the slugs don't end in `/`

I fixed those issues in the script, and fixed the broken links.

# Testing
- [x] Link checking detected new broken links
- [x] Output says it's still processing a ton of links, so I haven't completely broken it:
```
=== Validation Statistics ===
Total markdown files processed: 34
Total links/images processed: 879
  Valid: 117
  Invalid: 0
Total links with fragments processed: 34
  Valid links with fragments: 31
  Invalid links with fragments: 0
===============================
```

### DIFF
--- a/scripts/checkLinks.ts
+++ b/scripts/checkLinks.ts
@@ -118,14 +118,8 @@ function checkLinks(): void {
         return; // Skip external links
       }
 
-      const siteLinks = ['/install', '/images', '/profile'];
-      for (const siteLink of siteLinks) {
-        if (link.startsWith(siteLink)) {
-          return; // Skip site links
-        }
-      }
-      if (link.startsWith('/#') || link == '/') {
-          return; // Skip site links
+      if (!link.startsWith('/docs')) {
+        return; // Skip site links
       }
 
       // Resolve the link
--- a/scripts/checkLinks.ts
+++ b/scripts/checkLinks.ts
@@ -118,11 +118,14 @@ function checkLinks(): void {
         return; // Skip external links
       }
 
-      const siteLinks = ['/', '/install', '/images', '/profile'];
+      const siteLinks = ['/install', '/images', '/profile'];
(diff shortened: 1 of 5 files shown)


scripts/checkLinks.ts
```diff
@@ -124,11 +118,8 @@ function checkLinks(): void {
         return; // Skip external links
       }
 
-      const siteLinks = ['/install', '/images', '/profile'];
-      for (const siteLink of siteLinks) {
-        if (link.startsWith(siteLink)) {
-          return; // Skip site links
-        }
+      if (!link.startsWith('/docs')) {
+        return; // Skip site links
```
bfops, line 122: I feel like this probably breaks the "checking image URLs for validity" logic.. but we were explicitly skipping `/images` before anyway, so..?

cloutiertyler: Yes, we're not checking image links for validity because they are shipped with the website. There is a long-standing issue to implement images in this repo, but we haven't done it yet.

cloutiertyler (approved):
This LGTM

scripts/checkLinks.ts
bfops: okay great :+1: then this isn't a regression

PR #347, opened 2025-05-10
Remove double code-lang buttons

sdawka:
Nested the lang specific text into the overall codeblock instead of duplicating below.


bfops (approved):
LGTM, thank you!

PR #349, opened 2025-05-16
Fix TypeScript Quickstart wrong import

toqueteos:
This change alone doesn't fix the TypeScript Quickstart but helps.

This other PR https://github.com/clockworklabs/SpacetimeDB/pull/2747 fixes the codegen issues related to TypeScript imports too.


jsdt (approved):
Thanks for making this change!

